PR #1, opened 2024-03-14
Created initial b2ai script with file load and STFT

Rahul-Brito:
All the relevant changes are in b2ai_script.ipynb. 

- Right now I am hard-resampling audio to 16khz
- I am curious if the way I am storing the data in a dictionary is the best way or not ( {filename: {signal: signal, metadata: metadata}} )
- If so, I will make the scaled functions for the other features we want (thanks for the initial code Satra


satra:
can we make these python scripts and not notebooks?

satra:
also add a minimal requirements.txt

PR #13, opened 2024-03-19
Add scripts for speech-to-text using whisper and stt+forced alignment with whisperX

900miles:
Adds two functions for using Whisper or WhisperX to transcribe an audio file, and can perform speaker diarization and forced alignment of text output if using WhisperX.


fabiocat93:
hey @900miles do you mind importing the packages you use in your `process.py` file and adding them to the dependencies of the package?

900miles:
New commit should allow working directly with Audio objects. I've also added a requirements.txt but I've never really made one before so I'm not sure if I did it correctly.

satra:
instead of a requirements.txt just add it to the `pyproject.toml`

satra:
also perhaps change the filename to `speech2text`.

900miles:
Done and done!

PR #14, opened 2024-03-20
adding opensmile

Rahul-Brito:
in progress, added a function but need to have it:

- take in Audio
- to Processing_signal


src/b2aiprep/process.py
```diff
@@ -131,6 +131,19 @@ def resample_iir(audio: Audio, lowcut: float, new_sample_rate: int, order: int =
...
+    filename: 'str', feature_level: str = 'func'
+) -> torch.tensor: #feature_set: opensmile.FeatureSet = opensmile.FeatureSet.eGeMAPSv02
+    
+    smile = opensmile.Smile(
+        feature_set=opensmile.FeatureSet.eGeMAPSv02, #or path to conf: 'gemaps/eGeMAPSv02.conf'
+        feature_level=feature_level,
+        verbose=True,
+    )
+    
+    return smile.process_files(filename)
```
satra, line 144: could we adjust this to Smile's audio-based method so we can use the Audio object? 

also can we parameterize the config set?

src/b2aiprep/process.py
```diff
@@ -131,6 +131,19 @@ def resample_iir(audio: Audio, lowcut: float, new_sample_rate: int, order: int =
...
+
+
```
satra, line 146: can we also add to `to_features`?

Rahul-Brito:
agree on both points

Rahul-Brito:
Implemented the above @satra. Not sure the protocol to merge, can I merge my own PR?

PR #15, opened 2024-03-20
New functionalities

fabiocat93:
- Added voice conversion functionality (freevc)
- Added speech to text functionality (whisper)
- Added unit tests for VC and S2T
- Added CLI support for VC and S2T
- Enabled HF datasets as an output type for "to_features"

Still believe that we need a good discussion on APIs and interfaces


src/b2aiprep/process.py
```diff
@@ -166,6 +170,103 @@ def to_features(
...
+
+        :param model_name: Name of the model to be used for voice conversion.
+        :param use_gpu: Boolean indicating whether to use GPU for model computation.
+        """
+        use_gpu = torch.cuda.is_available()
+        self.tts = TTS(
+            model_name=model_name, 
+            progress_bar=progress_bar, 
+            gpu=use_gpu
+            )
```
satra, line 207: this is specific to cuda drivers, do you know if it supports `mps` on apple silicon?

src/tests/test_speech_to_text.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+from unittest.mock import patch, MagicMock
```
satra, line 2: i'm used to pytest over unittest as it has more features that make some components easier. could we switch to pytest here?

src/test_voice_converstion.py
```diff
@@ -0,0 +1,59 @@
+import unittest
```
satra, line 1: should this be moved to tests?

src/b2aiprep/process.py
```diff
@@ -166,6 +170,103 @@ def to_features(
...
+        # Perform voice conversion without modifying the source or target audio data directly.
+        with torch.no_grad():
+            self.tts.voice_conversion_to_file(source_wav=source_file, target_wav=target_file, file_path=output_file)
+
+
+
+
+
+
+class SpeechToText:
```
satra, line 231: how does this relate to miles' PR? could you please take a look?

src/b2aiprep/process.py
```diff
@@ -166,6 +170,103 @@ def to_features(
...
+    def __init__(
+            self, 
+            model_id: str = "openai/whisper-tiny",
+            max_new_tokens: int = 128, 
+            chunk_length_s: int = 30, 
+            batch_size: int = 16,
+            return_timestamps: Union[bool, str] = False,
+            ) -> None:
+        device = "cuda" if torch.cuda.is_available() else "cpu"
```
satra, line 240: let's allow for device to be set by user. on apple silicon, one can get hardware acceleration through `mps`.

satra (commented):
thanks @fabiocat93 - looks great. left a few comments to tweak a few things.

src/b2aiprep/process.py
fabiocat93: from my understanding, they started brainstorming movint to .to(device) but they never finished and mps is not supported right now for voice conversion

PR #17, opened 2024-03-25
added fig generation code

Rahul-Brito:
added figure generation code. The test figure I was getting looked weird so I need to maybe debug, but submitting PR for now. Code seems good?


src/b2aiprep/process.py
```diff
@@ -147,9 +150,10 @@ def extract_opensmile(
 
 def to_features(
     filename: Path,
-    subject: str,
-    task: str,
+    subject: str = None,
+    task: str = None,
     outdir: Path = Path(os.getcwd()),
+    save_figures: bool = True,
```
satra, line 156: should be added as an option to the CLI command.

src/b2aiprep/process.py
```diff
@@ -172,18 +176,45 @@ def to_features(
...
         "opensmile": features_opensmile,
         "sample_rate": audio.sample_rate,
         "checksum": md5sum,
     }
-    outfile = outdir / f"sub-{subject}_task-{task}_md5-{md5sum}_features.pt"
+    
+    #outfile = outdir / f"sub-{subject}_task-{task}_md5-{md5sum}_features.pt"
+    #saving file with just UUID name for now. TO DO: change it back possibly
+    feat_file_name = Path(filename).stem
+    outfile = outdir / f"{feat_file_name}_features.pt"
```
satra, line 196:
```suggestion
    if subject is not None:
        if task is not None:
            prefix = f"sub-{subject}_task-{task}_md5-{md5sum}"
        else:
            prefix = f"sub-{subject}_md5-{md5sum}"
    else:            
        prefix = Path(filename).stem
    outfile = outdir / f"{prefix}_features.pt"
```

src/b2aiprep/process.py
```diff
@@ -172,18 +176,45 @@ def to_features(
...
-    return features, outfile
+    
+    if save_figures:
+        #save spectogram as figure
+        
+        #general log spectrogram for image
+        features_specgram_log = specgram(audio, log=True)
+        
+        fig, ax = plt.subplots(figsize = (20,10))
+        ax.matshow(features_specgram_log, origin='lower', aspect=8, norm=LogNorm(vmin=0, vmax=1))
```
satra, line 207: already getting log specgram.

```suggestion
        ax.matshow(features_specgram_log, origin='lower', aspect=8)
```

src/b2aiprep/process.py
```diff
@@ -172,18 +176,45 @@ def to_features(
...
+        ax.axes.xaxis.set_ticks_position('bottom')
```
satra, line 208: need to add proper and x and y limits.

satra:
putting tags back so we can release when this is merged.

satra:
in addition to the CLI, which needs to take into account a list of files as opposed to file,subject,task, can you check/update the README so that alex's team can use this on their own with the new csv format?

src/b2aiprep/process.py
Rahul-Brito: ended up just putting in a better hard-coded image aspect that seems to work across speaking, breathing, coughing

src/b2aiprep/process.py
Rahul-Brito: think I did it; didn't test batch_convert since I wasn't 100% sure how to test the cli command from my dev branch

Rahul-Brito:
@satra i think this should be ready, maybe i could handle the paths in `batch_csv_convert()` as Path instead of str, and I did not actually run the CLI commands to test

satra:
since we don't have tests yet, please do run the CLI commands as that's what they would be using.

Rahul-Brito:
@satra should be ready to merge

PR #54, opened 2024-06-07
tie together the various constants used for parsing the redcap CSV

alistairewj:
I refactored the plethora of hard-coded dictionaries we had into a single `Enum` with an `Instrument` class. The goal of this was to (1) tie together all these constants so we wouldn't forget to add a `columns` variable for a questionnaire with the `name` variable, (2) reduce the number of hard-coded values, and (3) reduce the complexity of multiple dictionaries interacting.

* The `Instrument` class contains the hard-coded constants we need for each type of redcap instrument, e.g. the `session_id` which unique identifiers it, the name of the JSON file with the columns, and the name of the FHIR schema that was automatically converted from ReproSchema
* I fixed a few typos in the name of the questionnaires
* I now impute the `redcap_instrument` column in the loaded CSV (so I renamed this to be specific to loading a redcap CSV). This makes it easier for the later code to filter rows/columns together, since we can rely on the `redcap_instrument` column always having a value
* Cleaned up the interface to the dataset class a bit. Now the `find_*` methods return a list of `Path` objects, no more awkward dictionary stuff. This also fixed a bug where multiple sessions for a participant were merged together implicitly. It's a better signature in general IMO.
* Removed a lot of the now redundant `list_*` methods
* Updated the tutorial and added some code on loading audio


ibevers (commented):
This look a lot better! I'm glad that this also led to some bugs getting fixed

PR #59, opened 2024-06-12
Alistair/resource fixes

alistairewj:
Noticed that the snakeCase is not respected in a few files. Not sure how this snuck through as I thought I fixed it all! Added a test to check for it.


alistairewj:
@ibevers i fixed a few bugs and made sure the tests were passing

having `assert NotImplementedError` in tests is no fun, hides the real bugs :P

ibevers:
@alistairewj Ah, yes, I will avoid having `NotImplementedError` going forward. What do you think of having warnings instead? That way, we are less likely to forget about tests that need to be implemented, but we can still find the real bugs🐛

alistairewj:
Shall I merge?

alistairewj:
> I don't understand why precancerousLesions.json is deleted here, but I trust that you have a good reason

Yep! I deleted `precancerouseLesions.json` (note the additional 'e' as a typo). The original file `precancerousLesions.json` is still there.

PR #63, opened 2024-06-24
Remove TTS and convert voice feature

alistairewj:
Adds streamlit to the pyproject.toml as it is required now

This is redundant to #62 - but I'm testing the CI to see if requiring np <2 is explicitly required (streamlit says it doesn't support np >= 2 and it's unclear if any of our other packages do)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,9 +24,11 @@ jobs:
 
       - name: Install dependencies
         run: |
+          sudo apt-get -y install cmake libgoogle-perftools-dev libunwind-dev
           python -m pip install --upgrade pip setuptools
-          pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
+          pip install torch torchvision "torchaudio>=2.0.0" --index-url https://download.pytorch.org/whl/cpu
           pip install --no-cache-dir --editable=".[dev,tts]"
+
       - name: Test with pytest
         run: |
           pytest tests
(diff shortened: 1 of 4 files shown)


alistairewj:
Adding streamlit as a dependency and some experimentation exposed the following conflicts:

```
The conflict is caused by:
    b2aiprep 0+untagged.1.g5ba01ee depends on numpy==1.26.4
    streamlit 1.35.0 depends on numpy<2 and >=1.19.3
    transformers 4.39.3 depends on numpy>=1.17
    b2aiprep[dev,tts] 0+untagged.1.g5ba01ee depends on numpy==1.26.4
    tts 0.22.0 depends on numpy==1.22.0; python_version <= "3.10"
```

https://github.com/sensein/b2aiprep/actions/runs/9666011062/job/26664513604#step:4:368

After debugging, my view of the most reasonable solution was to remove the TTS package dependency. This means the CLI no longer supports converting a voice, which was a niche use anyway. The TTS package owners have recently shut down anyway: https://coqui.ai/

If we want to reimplement the feature, we probably need to use NeMo or something similar.

DebuggerDan:
Would `Torchvision` still be required as a dependency? ([merge commit](https://github.com/sensein/b2aiprep/commit/e3f0ceb06f0c773ac210c4573d68e2d9586e3b63))

Currently, when running `streamlit` using the README.md example & the dataset, the `b2aiprep-cli` program outputs `torchvision is not available - cannot save figures` - I am assuming this might be a remnant library initialization from the TTS & voice conversion features, but asking to be sure!

alistairewj:
Yes AFAIK we don't use any of the features from torchvision. This is reported by one of the imports, though not sure which one in particular.

PR #64, opened 2024-06-24
Audio tutorial

alistairewj:
This PR

- Adds `load_recording(recording_id: str)`, `load_recordings()`, and `load_spectrograms()` as functions to the dataset class
- Adds a jupyter notebook tutorial which loads in all the audio and transcribes a record

### DIFF
--- a/docs/tutorial_audio.ipynb
+++ b/docs/tutorial_audio.ipynb
@@ -1,0 +1,319 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Audio tutorial\n",
+    "\n",
+    "This tutorial walks through loading in individual audio records as well as the entire dataset."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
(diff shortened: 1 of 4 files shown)


ibevers (approved):
This looks good. It's convenient to be able to load all the recordings at once.

ibevers:
@alistairewj it looks like there is a path issue in the CI:
```
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_summer_school_data.py:6: in <module>
    from b2aiprep.summer_school_data import (
src/b2aiprep/summer_school_data.py:46: in <module>
    from b2aiprep.prepare import redcap_to_bids
src/b2aiprep/prepare.py:44: in <module>
    from b2aiprep.utils import _transform_str_for_bids_filename
E   ModuleNotFoundError: No module named 'b2aiprep.utils'
```

alistairewj:
Whoops, forgot to include that file. I moved the transform file to a separate module as it is imported from multiple modules, and it's easier to avoid circular imports that way

PR #71, opened 2024-06-27
Fix duplicate pd df indexes bug

ibevers:


alistairewj:
this is what I get for trying out new pandas functions without testing them first 😂

ibevers:
All good🙂

PR #74, opened 2024-07-02
Replace process.py with senselab functions

ibevers:
- [x] Audio (on SL)
- [x] embed_speaker (on SL) 
- [x] resample_iir (in review on SL)
- [x] verify_speaker (in review on SL)
- [x] verify_speaker_from_files (in review on SL)
- [x] specgram (on SL)
- [x] melfilterbank (on SL) 
- [x] MFCC (on SL) 
- [x] extract_opensmile (on SL)
- [x] plot_waveform (on SL)
- [x] plot_spectrogram (on SL)
- [x] plot_save_figure (deleted)
- [x] to_features (delete and replace usage with calls to individual functions)
- [x] SpeechToText (on SL)
- [x] transcribe (on SL)
- [x] to_hf_dataset (moved to one place since it is only used once in the CLI)


ibevers:
Changing to SenseLab functions might make breaking changes. Will try to replace functions with SenseLab usage to avoid this as much as possible.

PR #94, opened 2024-11-20
refactor derivative generation

satra:
refactor derivatives output keeping with proposed form


ibevers:
@satra I'll run this on what we have so far to check.

alistairewj (approved):

PR #101, opened 2024-12-09
100 refactor and tests for phenotype json code

ibevers:
Test phenotype JSON generation modules, move them out of notebook, and make them maintainable.


ibevers:
Updates 12/09:
Refactored and added more tests.

alistairewj:
> Updates 12/17: Pretty much done. Need to run whole thing manually once which takes ~10 minutes and check the output.
> 
> * [ ]  Decide how to integrate long-running test, so it is not run every time

One thing I would suggest is to have these tests triggered by the manual trigger on GitHub (I think it is `workflow_dispatch`). Then these tests will only be triggered when you manually click run on GitHub.

ibevers:
> > Updates 12/17: Pretty much done. Need to run whole thing manually once which takes ~10 minutes and check the output.
> > 
> > * [ ]   Decide how to integrate long-running test, so it is not run every time
> 
> One thing I would suggest is to have these tests triggered by the manual trigger on GitHub (I think it is `workflow_dispatch`). Then these tests will only be triggered when you manually click run on GitHub.

Will do!

ibevers:
> Updates 12/17: Pretty much done. Need to run whole thing manually once which takes ~10 minutes and check the output.
> 
> * [x]  Decide how to integrate long-running test, so it is not run every time
> * [ ]  Integrate into CLI

Decided to integrate as notebook rather than CLI command. May integrate into CLI later.

PR #110, opened 2025-01-06
Refactor commands.py

ibevers:
Ensures that only CLI specific code is in `commands.py` and all general-purpose code is in reusable modules.

### DIFF
--- a/src/b2aiprep/commands.py
+++ b/src/b2aiprep/commands.py
@@ -1,16 +1,18 @@
 """Commands available through the CLI."""
 
+# Refactor to improve readability, maintainability, and scalability
+
 import csv
-from functools import partial
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


satra:
@ibevers - can you describe what this is for? also does pre-commit not handle some of the stylistic components for this?

ibevers:
> @ibevers - can you describe what this is for? also does pre-commit not handle some of the stylistic components for this?

There is general purpose logic in `commands.py` that should be refactored into reusable modules. I feel that `commands.py` should have only code that is specific to the CLI interface, and everything else should be imported from general-purpose modules.

https://github.com/sensein/b2aiprep/blob/4ca121b1e755c6cede7a9a7abf05f92534b423bf/src/b2aiprep/commands.py#L333

PR #120, opened 2025-01-31
119 audioqc training

ibevers:


ibevers:
Depends on having a training dataset--can be done with bioacoustic qc

ibevers:
- [ ] decide which models to include in `models` directory

PR #124, opened 2025-02-10
Adding snr and l/h ratio to extract_single

Evan8456:
- Added functions to compute snr and l/h ratio


satra:
@Evan8456 - can you add these to senselab instead of here? there are also other tools that can help there with computing these.

ibevers:
@Evan8456 what's the status of this?

Evan8456:
@ibevers The Snr calculation isn't needed here anymore as I believe you added that to senselab in the most recent PR. I can  make a pr to senselab to add add L/H ratio.

ibevers:
@Evan8456 can you also add your SNR implementation? I think different SNRs vary wildly in their performance for different cases, so we should include different implementations!

PR #128, opened 2025-02-25
127 update setup.py and readme to ensure arm python on mac

ibevers:
Updates the README to inform macOS users that they need an ARM version of python.
Adds a check when installed on macOS to ensure that the python used is for ARM.


satra (approved):
overall looks good to me. did someone test this out?

ibevers:
Yes, it worked when I tested it:
Screenshot 2025-03-12 at 8 10 53 PM

It installs correctly when using my default (arm) python.

PR #141, opened 2025-03-27
Added update feature

Evan8456:
- Added `--update` feature to `prepare bids`. It creates a shadow copy of the bids folder and avoids re-generating already generated features.
- Added command `generate-audio-features` that allows us to generate features to a source destination, with the ability to avoid re-generating previous features.  
- Added `bids2shadow` that creates a shadow tree of the bids folder, that only contains the feature extractions (.pt files).


ibevers:
Note: @Evan8456 and I discussed making a separate create shadow tree function that simply copies all the pt files in the same directory structure with no other files. This should avoid copying all the files and deleting them and just copy the pt files in the same directory structure

ibevers:
@Evan8456 What does adding update give us that is not provided by overwrite? I imagine that it's pretty unlikely for there to be many incomplete .pt files, which is the only situation where update seems useful

Evan8456:
@ibevers  The overwrite flag, makes it so regardless of whether the .pt files were there already prior to feature extract, it re-extracts all the features. update is there in place to allow us to avoid re-extracting everything. It also has a use case for when we decide to add new features to senselab or b2aiprep and allow us to avoid any redundant extractions.

src/b2aiprep/commands.py
```diff
@@ -204,33 +205,24 @@ def prepare_bids(
...
-            subject_id=subject_id,
-        )
-    _LOGGER.info("Audio feature extraction complete.")
+    bids_path = Path(bids_dir_path)
+
+    generate_features_wrapper(
+        bids_path=bids_path,
+        transcription_model_size=transcription_model_size,
+        n_cores=n_cores,
+        with_sensitive = with_sensitive,
```
alistairewj, line 214: Look at turning on code linting, auto formatting on save, or something similar to avoid minor syntactically valid variations like the spacing here from cluttering the code base

src/b2aiprep/commands.py
```diff
@@ -1004,3 +1082,40 @@ def reproschema_to_redcap(audio_dir, survey_file, redcap_csv, participant_group)
 
     merged_csv_path = os.path.join(redcap_csv, "merged-redcap.csv")
     df_final.to_csv(merged_csv_path, index=False)
+
+
+@click.command()
+@click.argument("bids_src_dir", type=str)
+@click.argument("dest_dir", type=str)
+def bids2shadow(bids_src_dir, dest_dir):
```
alistairewj, line 1090: Shadow is not a good term tbh - I'd rename this to better reflect its function. copy_bids_derived_data (or similar) would be better

src/b2aiprep/prepare/prepare.py
```diff
@@ -124,19 +146,97 @@ def extract_single(
...
+                "formants": False,
+                "spectral_moments": False,
+                "pitch": False,
+                "slope_tilt": False,
+                "cpp_descriptors": False,
+                "duration": False,
+                "jitter": False,
+                "shimmer": False,
+            }
+            if not all(features in praat_features for features in FEATURE_EXTRACTION_SPEECH_RATE):
```
alistairewj, line 170: move all of these if statements out into a utility function, something like `create_praat_mouth_config(features)`

src/b2aiprep/prepare/prepare.py
```diff
@@ -124,19 +146,97 @@ def extract_single(
         "plugin": "serial",
     }
     # Extract features
+    if update and not overwrite:
```
alistairewj, line 149: hmmm do you need to keep checking update / overwrite? if you have loaded everything in `preloaded_pt`, can't you just carry on defining the configs based on the presence of features in `preloaded_pt`, and then write out the file. you've already verified that the user wants you to overwrite any existing data. the only check I can see doing again later is if the config is all False (i.e. all the features are the same / no updates to do), in which case you could return early to avoid writing out the same data to file

ibevers:
@Evan8456 when I run this, I get:
```
INFO:b2aiprep.commands:Beginning audio feature extraction...
INFO:b2aiprep.prepare.prepare:Number of audio files: 0
INFO:b2aiprep.commands:Audio feature extraction complete.
WARNING:b2aiprep.prepare.bids:Missing transcriptions for 18636 / 18636 audio files
WARNING:b2aiprep.prepare.bids:Missing features for 18636 / 18636 audio files
INFO:b2aiprep.commands:Process completed.
```

However, no features are generated:
```
(base) [isaacb@orcd-login001 ~]$ find /orcd/data/satra/002/datasets/b2aivoice/2.0-release/with-features/ -type f -name "*.pt"
```

sbatch script:
```
#!/bin/bash 
#SBATCH -c 4 # Cores per task
#SBATCH -n 24 # n tasks
#SBATCH -N 1 #Request 1 node
#SBATCH -t 48:00:00 #Max runtime
#SBATCH -p mit_preemptable 
#SBATCH --requeue
#SBATCH --mem-per-cpu=4G #Memory per core
#SBATCH -o b2ai-output/output_%j.txt #redirect output
#SBATCH -e b2ai-error/error_%j.txt #redirect errors
#SBATCH --mail-type=ALL #Mail for state changes
#SBATCH --mail-user=isaacb@mit.edu #email recipient
#SBATCH --nodelist=node1625

# Read the argument from args.txt using SLURM_ARRAY_TASK_ID
cd /orcd/data/satra/002/datasets/b2aivoice/2.0-release
source ~/miniconda3/bin/activate
conda activate b2ai
CACHEDIR="/home/isaacb/cache/sidx-${SLURM_ARRAY_TASK_ID}"
mkdir -p $CACHEDIR
rm $CACHEDIR/*.lock
PYTHONWARNINGS="ignore" 

b2aiprep-cli generate-audio-features \
/orcd/data/satra/002/datasets/b2aivoice/wasabi-exports/eipm-bridge2ai-internal-data-dissemination/2025-03-20T16.37.50.087Z/b2aiprep \
/orcd/data/satra/002/datasets/b2aivoice/2.0-release/with-features \
-t large-v3-turbo \
-n 24 \
--with_sensitive \
--validate \
--cache $CACHEDIR \
-s $ARG \
--update
```

Evan8456:
@ibevers is the folder you passed in already in bids format?

Evan8456:
Try running the folder with either prepare bids or redcap2bids and then generate features. The generate features command assumes the source directory is in bids already.

ibevers:
@Evan8456 I'm almost sure it's already in bids from inspection has all the subject dirs etc. Also, I know that Alex from TDOM generated this with prepare-bids

ibevers:
@Evan8456 the audio files are in the dir:
```
(base) [isaacb@orcd-login001 b2aiprep]$ pwd
/orcd/data/satra/002/datasets/b2aivoice/wasabi-exports/eipm-bridge2ai-internal-data-dissemination/2025-03-20T16.37.50.087Z/b2aiprep
(base) [isaacb@orcd-login001 b2aiprep]$ find . -type f -name "*.wav" | wc -l
18636
```

satra:
please use the `prepare-bids` command: https://github.com/sensein/b2aiprep/blob/93b9a1ed46b256cfd77a8c15b6e58d489f8ff2bb/src/b2aiprep/commands.py#L140

this will run without the input csv and audio data if pointed to an already generated bids dataset. you can take a look at the code internally to see what it's doing.

ibevers:
> please use the `prepare-bids` command:
> 
> https://github.com/sensein/b2aiprep/blob/93b9a1ed46b256cfd77a8c15b6e58d489f8ff2bb/src/b2aiprep/commands.py#L140
> 
> this will run without the input csv and audio data if pointed to an already generated bids dataset. you can take a look at the code internally to see what it's doing.

Okay

satra:
@Evan8456 - same question here. is this ready for merge?

PR #144, opened 2025-04-08
143 ensure create derived dataset works on release 2

ibevers:


satra:
i'll leave this open for now. i'm going to check if the redcap2bids already populates participants.tsv with all the relevant info (this is what it looks like). in this setting, it would simplify the derived dataset creation code.

PR #150, opened 2025-04-08
146 update for release 2

ibevers:


satra:
@Evan8456 - should there be an update to participants.json in the bids tree? it seems that is used to create the `participants.tsv` in the redcap_to_bids function.

Evan8456:
From what I understand, `participants.tsv` is created based off of participants.json. However, `participants.tsv ` seems to be a larger tsv with columns from various other tsvs from the phenotype folder. I noticed that we mainly use the phenotype tsvs for deriving the dataset in `create_derived_dataset()` rather than `participants.tsv`(the latter is only used to map record_ids to participant_ids). I'm not sure if it is necessary as all the data is captured in the phenotype, but I can update the json with all the remaining columns tomorrow.

PR #167, opened 2025-04-18
patch for release v2.0.0

alistairewj:
Small fixes to enable publishing v2.0.0

- trimmed participant ID and session ID to 8 characters as publishing platform only supports files up to 100 characters
    - and honestly having filenames > 100 characters and participant IDs > 32 characters is pretty tedious anyway
    - this has a 1% chance of a conflict once we hit 10,000 participants so the code will need updating before then
-  fixes the participant ID removal to apply to all derived/published datasets, before it missed the derived dataset
- clean up the README and data description
- copy over the fixed resources from this repo (readme, data description, changelog) to the published dataset within the publish command

### DIFF
--- a/src/b2aiprep/commands.py
+++ b/src/b2aiprep/commands.py
@@ -1,5 +1,6 @@
 """Commands available through the CLI."""
 
+from copy import deepcopy
 import csv
 import json
 import logging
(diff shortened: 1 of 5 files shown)


ibevers:
@alistairewj Did you run this and verify they work? Do I need to do that?

alistairewj:
I think independently running it can't hurt, but yes I did have it running myself

ibevers:
@alistairewj Okay, I will run it on Engaging, and will merge if there are no issues

PR #179, opened 2025-05-30
Fix/issue#174

Evan8456:
- Fixing issue with new version of senselab not working with b2aiprep

### DIFF
--- a/src/b2aiprep/prepare/prepare.py
+++ b/src/b2aiprep/prepare/prepare.py
@@ -111,8 +111,7 @@ def extract_single(
 
     logging.disable(logging.ERROR)
     # Load audio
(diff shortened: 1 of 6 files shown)


src/b2aiprep/prepare/prepare.py
```diff
@@ -111,8 +111,7 @@ def extract_single(
 
     logging.disable(logging.ERROR)
     # Load audio
-    audio_orig = Audio.from_filepath(wav_path)
```
ibevers, line 114: @Evan8456 do you mind updating the other ones too?
Screenshot 2025-06-04 at 11 00 38 AM

ibevers (approved):
Given our limited test coverage, I'm nervous that this will break something, but we can fix those issues as they come up.

PR #183, opened 2025-06-25
181 add at least one test for each function

ibevers:
- Adds tests with the goal of ensuring there is at least one test per function.
- Updates bids.py based on the tests.


ibevers (commented):
I'm okay with this. Do you see anything that needs to be changed @Evan8456?

Evan8456:
lgtm!

PR #195, opened 2025-09-19
Initial commit of peds id remapping

Evan8456:
- Added script to generate id remapping
- Would need to save the look up table that is generated.


external_scripts/peds_id_remap.py
```diff
@@ -0,0 +1,96 @@
...
+        id_column = "record_id"
+    elif "participant_id" in df.columns:
+        id_column = "participant_id"
+    else:
+        raise ValueError(
+            "No valid ID column found (expected 'record_id' or 'participant_id')"
+        )
+
+    id_list = list(set(df[id_column].tolist()))
+    id_list = sorted(id_list, key=lambda x: (int(x[:3]), x[3:]))
```
wilke0818: what is happening here? the first 3 vs. the rest are being used how/why?

wilke0818 (approved):
nothing necessarily needs changes. just some better practices I'd recommend

PR #201, opened 2025-11-14
Made fixes to redcap import script & publish command

Evan8456:
- Updated redcap import script to be compatible with b2ai redcap v4.3.0
- Updated publish to maintain bids format and renamed it to deidentify_bids_dataset
- Fixed feature extraction logic to always generate features to target directory

### DIFF
--- a/src/b2aiprep/prepare/dataset.py
+++ b/src/b2aiprep/prepare/dataset.py
@@ -676,15 +676,17 @@ class BIDSDataset:
         # sub-<participant_id>_ses-<session_id>_task-<task_name>_run-_metadata.json
         filename = f"sub-{subject_id}"
         if session_id is not None:
-            session_id = session_id.replace(" ", "-")
+            session_id = session_id.replace(" ", "-").replace("_", "-")
             filename += f"_ses-{session_id}"
         if task_name is not None:
             task_name = task_name.replace(" ", "-")
+            task_name = task_name.replace("_", "-")
             if recording_name is not None:
                 task_name = recording_name.replace(" ", "-")
+                task_name = recording_name.replace("_", "-")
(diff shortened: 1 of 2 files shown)


src/b2aiprep/prepare/dataset.py
```diff
@@ -1425,7 +1425,7 @@ def _deidentify_audio_files(
             # Create output path with deidentified structure
             audio_path_stem_ending = "_".join(audio_path.stem.split("_")[2:])
             output_path = outdir.joinpath(
-                f"sub-{participant_id}/ses-{session_id}/audio/{audio_path_stem_ending}.wav"
+                f"sub-{participant_id}/ses-{session_id}/audio/sub-{participant_id}_ses-{session_id}_{audio_path_stem_ending}.wav"
```
wilke0818: @alistairewj can you confirm that we're ok to have the BIDs like structure here? I believe from your emails it was originally removed due to PhysioNet length restrictions.

wilke0818 (approved):
I just want @alistairewj to confirm we won't shoot ourselves in the foot with returning to the BIDS filename format with PhysioNet but otherwise looks reasonable. All comments are perhaps just suggestions for an ideal case for the package

src/b2aiprep/prepare/dataset.py
alistairewj: our participant IDs and session IDs for pediatrics are now short enough that it works

I haven't verified this for the adult data but anything 8 characters or less will be fine

alistairewj (changes requested):
the derived command still looks to be doing deidentification, e.g. calculating sensitive audio, filtering audio files. all deidentification should be done at the earlier step

PR #207, opened 2025-12-02
Reorganize phenotype files

alistairewj:
Now uses a submodule for b2ai-redcap2rs -> loads phenotype organization directly from redcap -> uses a custom maintained CSV file to decide the ultimate phenotype structure.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -200,11 +200,15 @@ def test_redcap2bids_cli(setup_temp_files):
     """Test the 'b2aiprep-cli redcap2bids' command using subprocess."""
     redcap_csv_path, audio_dir, _, _ = setup_temp_files
 
+    project_root = Path(__file__).parent.parent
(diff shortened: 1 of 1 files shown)


Evan8456:
@alistairewj The test cases are failing because we added a new submodule. The tests fail because they aren’t being run from the directory where the b2aiprep folder is located, so the reproschema files can’t be found.

src/b2aiprep/commands.py
```diff
@@ -558,6 +545,61 @@ def validate_data(derivatives_csv_path):
...
+    type=click.Path(path_type=Path, exists=True, resolve_path=True),
+    default=None,
+    help="Optional path to the protocol schema file to parse.",
+)
+@click.option(
+    "--reorganize",
+    is_flag=True,
+    help="Reorganize phenotype JSONs into manually curated file organization."
+)
+@click.option(
```
wilke0818: should be optional no?

wilke0818 (approved):

src/b2aiprep/commands.py
alistairewj: it is!

PR #215, opened 2025-12-11
Updates for release v3.0

alistairewj:
Many updates - this code is used for the (ever hopeful) final run to generate v3.0.

- Cleaned up a lot of old code
- Fixed pediatric session IDs and task names post-hoc (this should be updated to fix session IDs and task names in the initial conversion)
- Multithreading for deidentify writing out files to speed it up significantly
- Session IDs are now output as 8 characters 0-9a-f, mainly because we don't have a way to deterministically sort session IDs, so it's safer to trim
- Fixed a bug where many phenotypes were filtered to only 1 row / participant
- Bundled data is now fp16 to save space
- Bundle features are output in a flat features/ folder rather than nested into very small group folders

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,15 +1,17 @@
 cff-version: 1.2.0
 message: "If you use this software, please cite it as below."
 authors:
-- family-names: "Bevers"
-  given-names: "Isaac"
-  orcid: "https://orcid.org/0009-0002-0074-0087"
-- family-names: "Ghosh"
-  given-names: "Satrajit"
-  orcid: "https://orcid.org/0000-0002-5312-6729"
 - family-names: "Johnson"
   given-names: "Alistair"
   orcid: "https://orcid.org/0000-0002-8735-3014"
+- family-names: "Bevers"
+  given-names: "Isaac"
+  orcid: "https://orcid.org/0009-0002-0074-0087"
+- family-names: "Ng"
+  given-names: "Evan"
+  orcid: "https://orcid.org/0009-0006-8554-5414"
+- family-names: "Wilke"
+  given-names: "Jordan"
 - family-names: "Brito"
   given-names: "Rahul"
   orcid: "https://orcid.org/0000-0003-0343-0455"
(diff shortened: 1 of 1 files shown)


src/b2aiprep/prepare/resources/bids_field_organization.csv
wilke0818: just going to trust you on this one

src/b2aiprep/commands.py
```diff
@@ -389,7 +387,7 @@ def create_bundled_dataset(bids_path, outdir, skip_audio, skip_audio_features):
     static_features = []
     for filepath in tqdm(audio_paths, desc="Loading static features", total=len(audio_paths)):
         pt_file = filepath.parent.joinpath(f"{filepath.stem}_features.pt")
-        if not filepath.exists(): #pt will exist based on how audio paths were generated, but audio file might not exist
+        if not pt_file.exists(): #pt will exist based on how audio paths were generated, but audio file might not exist
```
wilke0818: not sure why this caused static feature generation to fail unless you didn't have the audio files in the folder you were bundling. This check only really does something if the features filename is malformed.

src/b2aiprep/commands.py
```diff
@@ -474,11 +470,11 @@ def create_bundled_dataset(bids_path, outdir, skip_audio, skip_audio_features):
         if len(ds) == 0:  
             _LOGGER.warning(
                 "No non-NaN entries found for %s feature. Skipping parquet export.",
-                f"{feature_class}_{feature_name}" if feature_class else feature_name
+                feature_alias
             )
             continue
         ds.to_parquet(
-            str((f"{feature_output_dir}/{feature_name}.parquet")),
+            features_dir.joinpath(f"{feature_alias}.parquet").resolve().as_posix(),
```
wilke0818: NIT: I think having folders within features per "source" of the feature is a bit nicer of organization compared to this (e.g. torchaudio/spectrograms.parquet vs. torchaudio_spectrograms.parquet)

tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.fixture(scope="module")
+def reproschema_module_path():
+    project_root = Path(__file__).parent.parent
+    reproschema_path = project_root.joinpath("b2ai-redcap2rs", "b2ai-redcap2rs").resolve().as_posix()
+    return reproschema_path
+
+
+
+@pytest.fixture
+def setup_publish_config(tmp_path):
```
wilke0818: pretty minor but can we can call it `setup_deidentify_config`

tests/test_bids_deidentification.py
```diff
@@ -133,16 +125,12 @@ def output_dir(self):
         # Cleanup
         shutil.rmtree(temp_dir)
 
-    def test_deidentify_basic_functionality(self, temp_bids_dir, output_dir):
+    def test_deidentify_basic_functionality(self, temp_bids_dir, output_dir, setup_publish_config):
```
wilke0818: and replace these: `setup_deidentify_config`

PR #234, opened 2026-01-22
Fix age on a separate row in phenotype

alistairewj:
The actual fix would be to split up the RedCap CSV into a single dataframe per repeat instrument, and then merge those together as needed.

For now I just did a patch to carry forward the age.

Also added a command for counting sessions in the RedCap CSV.


wilke0818 (approved):
looks good. runs locally. Doesn't perfectly match PhysioNet but the differences actually make this method seem preferable to those that were used to deploy PhysioNet as some of the ages there are NaNs where less (if any) are now.

PR #256, opened 2026-03-30
Added pii detection command

Evan8456:
- Added new command to run presidio pii detection model on transcripts to help check for pii

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
     "presidio-anonymizer",
     "transformers>=4.40.0",
     "accelerate>=0.27.0",
+    "gliner>=0.2.0",
     "huggingface-hub>=0.22.0",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,9 @@ dependencies = [
     "pydra",
     "torchaudio~=2.8.0",
     "torch~=2.8.0",
-    "torchvision~=0.23.0"
+    "torchvision~=0.23.0",
+    "presidio-analyzer",
+    "presidio-anonymizer",
 ]
 
 [tool.setuptools.packages.find]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,13 +33,16 @@ dependencies = [
     "torchvision~=0.23.0",
     "presidio-analyzer",
     "presidio-anonymizer",
+    "transformers>=4.40.0",
+    "accelerate>=0.27.0",
+    "huggingface-hub>=0.22.0",
 ]
 
 [tool.setuptools.packages.find]
-where = ["src"] # <-- this changed
+where = ["src"]
 
 [tool.setuptools.package-data]
-"*" = ["*.*", ".gitkeep"] # <-- this changed
+"*" = ["*.*", ".gitkeep"]
 
 [project.optional-dependencies]
 doc = [
@@ -60,6 +63,13 @@ dashboard = [
     "b2aiprep[doc]",
     "streamlit>=1.42.2",
     "senselab[audio]"
+]
+# Optional: only install when needed (heavy download ~8GB model)
+llm = [
+    "transformers>=4.40.0",
+    "accelerate>=0.27.0",
+    "huggingface-hub>=0.22.0",
+    "bitsandbytes>=0.43.0",  # for 4-bit quantization
 ]
 
 [project.scripts]
--- a/src/b2aiprep/cli.py
+++ b/src/b2aiprep/cli.py
@@ -22,8 +22,6 @@ from b2aiprep.commands import (  # gensynthtabdata,
     run_quality_control_on_audios,
     update_bids_template_command,
     pii_detection,
-    pii_detection_phi4,
-    pii_detection_gliner,
     task_correctness_phi4,
 )
 
@@ -67,8 +65,6 @@ cli.add_command(validate_feature_extraction)
 cli.add_command(run_quality_control_on_audios)
 cli.add_command(update_bids_template_command)
 cli.add_command(pii_detection)
-cli.add_command(pii_detection_phi4)
-cli.add_command(pii_detection_gliner)
 cli.add_command(task_correctness_phi4)
 
 if __name__ == "__main__":
--- a/src/b2aiprep/cli.py
+++ b/src/b2aiprep/cli.py
@@ -23,6 +23,8 @@ from b2aiprep.commands import (  # gensynthtabdata,
     update_bids_template_command,
     pii_detection,
     pii_detection_phi4,
+    pii_detection_gliner,
+    task_correctness_phi4,
 )
 
 
@@ -66,6 +68,8 @@ cli.add_command(run_quality_control_on_audios)
 cli.add_command(update_bids_template_command)
 cli.add_command(pii_detection)
 cli.add_command(pii_detection_phi4)
+cli.add_command(pii_detection_gliner)
+cli.add_command(task_correctness_phi4)
 
 if __name__ == "__main__":
     # include main to enable python debugging
--- a/src/b2aiprep/cli.py
+++ b/src/b2aiprep/cli.py
@@ -21,6 +21,7 @@ from b2aiprep.commands import (  # gensynthtabdata,
     deidentify_bids_dataset,
     run_quality_control_on_audios,
     update_bids_template_command,
+    pii_detection,
 )
 
 
@@ -62,6 +63,7 @@ cli.add_command(bids2shadow)
 cli.add_command(validate_feature_extraction)
 cli.add_command(run_quality_control_on_audios)
 cli.add_command(update_bids_template_command)
+cli.add_command(pii_detection)
 
 if __name__ == "__main__":
     # include main to enable python debugging
--- a/src/b2aiprep/cli.py
+++ b/src/b2aiprep/cli.py
@@ -22,6 +22,7 @@ from b2aiprep.commands import (  # gensynthtabdata,
     run_quality_control_on_audios,
     update_bids_template_command,
     pii_detection,
+    pii_detection_phi4,
 )
 
 
@@ -64,6 +65,7 @@ cli.add_command(validate_feature_extraction)
 cli.add_command(run_quality_control_on_audios)
 cli.add_command(update_bids_template_command)
 cli.add_command(pii_detection)
+cli.add_command(pii_detection_phi4)
 
 if __name__ == "__main__":
     # include main to enable python debugging
--- a/src/b2aiprep/commands.py
+++ b/src/b2aiprep/commands.py
@@ -1528,7 +1528,7 @@ def pii_detection(bids_folder, outdir):
 
     for path in tqdm(paths):
         pt_dict = torch.load(path, map_location="cpu", weights_only=False)
-        transcript = str(pt_dict["transcription"])
+        transcript = pt_dict.get("transcription")
 
         if transcript is None:
             _LOGGER.warning(f"No 'transcription' key in {path}, skipping.")
--- a/src/b2aiprep/commands.py
+++ b/src/b2aiprep/commands.py
@@ -69,7 +69,7 @@ from b2aiprep.prepare.utils import normalize_task_label
 from b2aiprep.prepare.update import TemplateUpdateError, reorganize_bids_activities, update_bids_template_files
 
 #from b2aiprep.prepare.models import prompt_gliner, prompt_phi4, prompt_presidio
-from b2aiprep.prepare.models import Models
+from b2aiprep.prepare.models import ModelRegistry
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -1544,7 +1544,7 @@ def pii_detection(bids_folder, outdir, model_choice):
     bids_folder = Path(bids_folder)
     outdir = Path(outdir)
     shutil.copytree(bids_folder, outdir)
-    model = Models()
+    model = ModelRegistry()
 
     paths = [f for f in outdir.rglob("*.pt")]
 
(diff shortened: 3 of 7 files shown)


src/b2aiprep/prepare/resources/few_shot_prompt.txt
wilke0818, line 1: is this the prompt that FHS used?

wilke0818: I'd have additionally expected a segment that specifies uniquely identifying information.

src/b2aiprep/cli.py
```diff
@@ -62,6 +66,10 @@ def cli(ctx, log_level):
 cli.add_command(validate_feature_extraction)
 cli.add_command(run_quality_control_on_audios)
 cli.add_command(update_bids_template_command)
+cli.add_command(pii_detection)
+cli.add_command(pii_detection_phi4)
+cli.add_command(pii_detection_gliner)
```
wilke0818: should probably be just one command? or not even a command but just existing functions that we later break out into a pipeline command

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+            outputs = model.generate(
+                inputs,
+                max_new_tokens=500,
+                temperature=0.7,
+                do_sample=True,
+                pad_token_id=tokenizer.eos_token_id,
+            )
+
+        llm_output = tokenizer.decode(outputs[0][inputs.shape[-1]:], skip_special_tokens=True)
+        pt_dict["pii_detection"] = llm_output
```
wilke0818: do we want to just save the llm output as is? seems like something we should verify in some way (parseable json) and generally we should have each of these pii functions create the same general output object if possible

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+        torch.save(pt_dict, path)
+
+@click.command()
+@click.argument("bids_folder", type=click.Path(exists=True))
+@click.argument("outdir", type=click.Path())
+def pii_detection_gliner(bids_folder, outdir):
```
wilke0818: can probably just combine into one PII function

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+        pt_dict = torch.load(path, map_location="cpu", weights_only=False)
+        transcript = pt_dict.get("transcription")
+
+        if transcript is None:
+            _LOGGER.warning(f"No 'transcription' key in {path}, skipping.")
+            continue
+        
+        model = GLiNER.from_pretrained("nvidia/gliner-pii")
+        pii_json = model.predict_entities(transcript, hippa_labels, threshold=0.5)
+        pt_dict["pii_detection"] = pii_json
```
wilke0818: currently putting it in the exact same place will override it if you wanted to run multiple of these models

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+            continue
+    
+        pt_dict = torch.load(path, map_location="cpu", weights_only=False)
+        transcript = pt_dict.get("transcription")
+
+        if transcript is None:
+            _LOGGER.warning(f"No 'transcription' key in {path}, skipping.")
+            continue
+
+        #change this
```
wilke0818, line 1636: agreed

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+            outputs = model.generate(
+                inputs,
+                max_new_tokens=500,
+                temperature=0.7,
+                do_sample=True,
+                pad_token_id=tokenizer.eos_token_id,
+            )
+
+        llm_output = tokenizer.decode(outputs[0][inputs.shape[-1]:], skip_special_tokens=True)
+        pt_dict["task_correctness"] = llm_output
```
wilke0818, line 1640: some form of llm output parsing should be considered

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+    outdir = Path(outdir)
+    shutil.copytree(bids_folder, outdir)
+    paths = [f for f in outdir.rglob("*.pt")]
+    
+    audio_task_path = files("b2aiprep.prepare.resources").joinpath("audio_task_descriptions.json")
+    with open(audio_task_path , 'r') as file:
+        audio_task = json.load(file)
+        
+    
+    model_id = "microsoft/Phi-4-mini-instruct"
```
wilke0818: why this model for checking correctness? I think it would be better to allow the user to pass in a model and default to one if they don't. Additionally, we should probably allow ollama models since I don't know if HFModels are as general purpose

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1505,266 @@ def redcap_stats(filename, num_sessions):
...
+
+    tokenizer = AutoTokenizer.from_pretrained(model_id)
+    model = AutoModelForCausalLM.from_pretrained(
+            model_id,
+            torch_dtype=torch.float16 if torch.cuda.is_available() else torch.float32,
+            device_map="auto" if torch.cuda.is_available() else None,
+        )
+
+    for path in tqdm(paths):
+        task = path.stem.split("task-")[-1]
```
wilke0818: this presumably will fail silently because paths are feature files, all of which are _task-TASK-NAME_features

pyproject.toml
```diff
@@ -30,14 +30,20 @@ dependencies = [
...
     "torchaudio~=2.8.0",
     "torch~=2.8.0",
-    "torchvision~=0.23.0"
+    "torchvision~=0.23.0",
+    "presidio-analyzer",
+    "presidio-anonymizer",
+    "transformers>=4.40.0",
+    "accelerate>=0.27.0",
+    "gliner>=0.2.0",
+    "huggingface-hub>=0.22.0",
```
wilke0818, line 39: more just surprised this one is needed since senselab i believe already uses this as a dependency

src/b2aiprep/prepare/resources/few_shot_prompt.txt
Evan8456: Yes, the prompt is the same as the FHS uses.

src/b2aiprep/cli.py
Evan8456: My issue is we're still not sure which model is best for for this. I can move them to functions add let the user specify.

src/b2aiprep/commands.py
Evan8456: Same as above.

src/b2aiprep/commands.py
Evan8456: Once I combine it into a single function, it should be not be an issue.

src/b2aiprep/commands.py
Evan8456: Place holder for now as we don't yet have a sense what what the ideal prompt is.

src/b2aiprep/commands.py
Evan8456: I'll do a `replace("_features", "")` for this.

src/b2aiprep/commands.py
```diff
@@ -1550,149 +1554,39 @@ def pii_detection(bids_folder, outdir):
 
     for path in tqdm(paths):
         pt_dict = torch.load(path, map_location="cpu", weights_only=False)
-        transcript = pt_dict.get("transcription")
+        transcript = str(pt_dict.get("transcription"))
```
wilke0818: if transcript is None/not there, this will create a string with "None". Python is weird

src/b2aiprep/commands.py
```diff
@@ -1719,52 +1613,28 @@ def task_correctness_phi4(bids_folder, outdir):
...
-        task = path.stem.split("task-")[-1]
+        task = path.stem.split("task-")[-1].replace("_features", "")
         instructions = audio_task.get(task, {}).get("instructions")
         if instructions:
             _LOGGER.warning(f"No 'instructions' found for task {task}, skipping.")
             continue
     
         pt_dict = torch.load(path, map_location="cpu", weights_only=False)
-        transcript = pt_dict.get("transcription")
+        transcript = str(pt_dict.get("transcription"))
```
wilke0818: same as above

tests/test_cli.py
wilke0818, line 1: so are we just not testing these at all now?

src/b2aiprep/commands.py
```diff
@@ -1719,52 +1613,28 @@ def task_correctness_phi4(bids_folder, outdir):
...
-            torch_dtype=torch.float16 if torch.cuda.is_available() else torch.float32,
-            device_map="auto" if torch.cuda.is_available() else None,
-        )
-
     for path in tqdm(paths):
-        task = path.stem.split("task-")[-1]
+        task = path.stem.split("task-")[-1].replace("_features", "")
         instructions = audio_task.get(task, {}).get("instructions")
         if instructions:
             _LOGGER.warning(f"No 'instructions' found for task {task}, skipping.")
```
wilke0818: get the path not the task as every task should have instructions (right?) so that is actually bad (maybe errorable)?

src/b2aiprep/prepare/models.py
wilke0818, line 1: maybe something more descriptive than models, unless you want to have all "models" that we end up making live in this one file. mostly worried about conflicts that could end up arising from more standard packages already using models

src/b2aiprep/prepare/models.py
```diff
@@ -0,0 +1,52 @@
...
+                max_new_tokens=500,
+                temperature=0.7,
+                do_sample=True,
+                pad_token_id=tokenizer.eos_token_id,
+            )
+
+        llm_output = tokenizer.decode(outputs[0][inputs.shape[-1]:], skip_special_tokens=True)
+        return llm_output
+
+    def prompt_presidio(self, prompt):
```
wilke0818: i understand why you named them these but given you're not prompting anything, doesn't really make sense

src/b2aiprep/prepare/models.py
```diff
@@ -0,0 +1,52 @@
...
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from presidio_analyzer import AnalyzerEngine
+from presidio_anonymizer import AnonymizerEngine
+from presidio_anonymizer.entities import OperatorConfig
+
+import torch
+
+class Models:
+    def __init__(self):
+        self.analyzer = AnalyzerEngine()
```
wilke0818, line 11: why have these in the init if they are only used for presidio

src/b2aiprep/prepare/models.py
```diff
@@ -0,0 +1,52 @@
+from gliner import GLiNER
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from presidio_analyzer import AnalyzerEngine
+from presidio_anonymizer import AnonymizerEngine
+from presidio_anonymizer.entities import OperatorConfig
+
+import torch
+
+class Models:
```
wilke0818: I am being an ass and nit-picky but I just don't like it

wilke0818 (changes requested):
extremely NIT comments so up to you for most. one bug needs fixing

tests/test_cli.py
Evan8456: There are tests located in `test_cli.py`

src/b2aiprep/prepare/models.py
Evan8456: Originally they were in the presdio function, but since we're looping over all files, it would create an AnalyzerEngine every time we want to prompt a file, which is inefficient.

wilke0818: Doing something similar to senselab where we generate models only once by creating a pseudo cache is probably the best move, but I don't think these classes will bog performance so good to leave it

src/b2aiprep/prepare/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+import torch
+
+class ModelRegistry:
+    def __init__(self):
+        self.analyzer = AnalyzerEngine()
+        self.anonymiser = AnonymizerEngine()
+
+    def prompt_phi4(self, prompt):
+        print(type(prompt))
```
wilke0818: remove (or log)

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1508,132 @@ def redcap_stats(filename, num_sessions):
         else:
             for record_id, n_sessions in over_n_sessions.items():
                 click.echo(f"Record ID: {record_id}, Sessions: {n_sessions}")
+
+@click.command()
+@click.argument("bids_folder", type=click.Path(exists=True))
+@click.argument("outdir", type=click.Path())
+@click.option("model_choice", "--m", type=str, default="presidio")
```
wilke0818: should be `@click.option("--model-choice", "-m", type=str, default="presidio")`

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1508,132 @@ def redcap_stats(filename, num_sessions):
...
+     Example updated .pt contents:
+        {
+            "transcript": <transcript>,
+            "task_correctness": "True"
+        }    
+    
+    """
+    bids_folder = Path(bids_folder)
+    outdir = Path(outdir)
+    shutil.copytree(bids_folder, outdir)
```
wilke0818: this will possibly error if they are the same directory, which if you were just updating the PT files would probably be how I would use it

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1508,132 @@ def redcap_stats(filename, num_sessions):
...
+    audio_task_path = files("b2aiprep.prepare.resources").joinpath("audio_task_descriptions.json")
+    with open(audio_task_path , 'r') as file:
+        audio_task = json.load(file)
+    
+    if not paths:
+        _LOGGER.warning(f"No .pt files were found in {bids_folder}")
+        return
+
+    for path in tqdm(paths):
+        task = path.stem.split("task-")[-1].replace("_features", "")
```
wilke0818: there might be one malformed peds task name that includes `task-` if I remmeber correctly. Change to `_task-` to be safe

src/b2aiprep/commands.py
```diff
@@ -1499,3 +1508,132 @@ def redcap_stats(filename, num_sessions):
...
+                    "score": 0.85
+                }
+            ],
+            "redacted_text": "Hi, my name is <PERSON> and my SSN is <US_SSN>."
+        }    
+    
+    """
+    bids_folder = Path(bids_folder)
+    outdir = Path(outdir)
+    shutil.copytree(bids_folder, outdir)
```
wilke0818: same as below - might error if they are same directory

wilke0818 (changes requested):

PR #302, opened 2026-04-15
Made id remapping into a command

Evan8456:
Made the id remapping a command so TDOM will know how to use it for future releases.

### DIFF
--- a/src/b2aiprep/commands.py
+++ b/src/b2aiprep/commands.py
@@ -59,7 +59,7 @@ from b2aiprep.prepare.prepare import (
 from b2aiprep.prepare.quality_control import quality_control_wrapper
 
 from b2aiprep.prepare.data_validation import validate_phenotype, validate_no_extra_audio_tasks_present
-from b2aiprep.prepare.utils import normalize_task_label
+from b2aiprep.prepare.utils import normalize_task_label, generate_or_load_seed, generate_pseudonym, load_lookup_table, build_lookup_table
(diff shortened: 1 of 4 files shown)


src/b2aiprep/commands.py
```diff
@@ -1499,3 +1499,39 @@ def redcap_stats(filename, num_sessions):
         else:
             for record_id, n_sessions in over_n_sessions.items():
                 click.echo(f"Record ID: {record_id}, Sessions: {n_sessions}")
+
+@click.command()
+@click.argument("input_file", type=click.Path(exists=True))
+@click.argument("output_dir", type=click.Path())
+@click.option("--load_lookup", type=click.Path(exists=True), default=None, show_default=True)
+def id_remap(input_file, output_dir, load_lookup):
```
wilke0818, line 1507: Minor: can you add some logs of how many previous IDs there were and how many there are now?

tests/test_cli.py
```diff
@@ -886,3 +886,29 @@ def test_reproschema_to_redcap_cli():
         result = subprocess.run(command, capture_output=True, text=True)
         assert result.returncode == 0, f"CLI command failed: {result.stderr}"
         assert output_dir.exists(), "Output directory was not created"
+
+
+
+def test_id_remap_cli():
```
wilke0818, line 892: code looks good. maybe one quick addition to this test/add another test that shows updating an id mapping works correctly (want to check that the old ID maps to the same thing after and that the new ID is in there).

wilke0818 (approved):

wilke0818 (approved):

PR #307, opened 2026-05-13
Redcap import

Evan8456:
Updated redcap import scripts


external_scripts/questionnaires/questionnaire.py
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    "record_id": id,
+                    "selected_language": "1",
+                    "consent_method": "paper",
+                    "enrolled": "1",
+                    "consent_date": consent_date,
+                    "is_feasibility_participant": "no",
+                    "enrollment_institution": "sickkids",
+                    "is_control_participant": "no",
+                    f"eligible_studies___{eligible_studies}": "1",
```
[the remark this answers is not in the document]
Evan8456: Age should never be <2, so this shouldn't be an issue.

external_scripts/questionnaires/questionnaire.py
```diff
@@ -0,0 +1,79 @@
...
+                }
+            ]
+        )
+        df = pd.concat([df, feas], ignore_index=True)
```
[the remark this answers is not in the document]
wilke0818: seems like a simple change for efficiency but also ~800 records max so prob wont matter?

external_scripts/questionnaires/questionnaire.py
```diff
@@ -0,0 +1,98 @@
...
+        "recording_name",
+        "recording_duration",
+        "recording_size",
+        "recording_profile_name",
+        "recording_profile_version",
+        "recording_microphone",
+        "participant_study_id"
+    ]
+
+    df.drop(columns=[c for c in columns_list if c in df.columns], axis=1, inplace=True)
```
wilke0818, line 134: why we dropping these?

external_scripts/questionnaires/questionnaire.py
wilke0818, line 1: why is it called feas

external_scripts/questionnaires/questionnaire-example.sh
wilke0818, line 1: Are these meant to be example bash scripts for tdom? If so I would just make a Markdown doc of how to use

external_scripts/recordings_redcap/recording.py
```diff
@@ -0,0 +1,41 @@
+import uuid
+import pandas as pd
+import argparse
+def create_recording_redcap(recording_csv, output_csv):
+    # recording_csv = (
```
wilke0818: delete

external_scripts/recordings_redcap/recording.py
```diff
@@ -0,0 +1,41 @@
...
+
+    l += [
+        "peds_mc_neck_mass___thyroglossal_duct_cyst",
+        "peds_mc_neck_mass___branchial_cleft_cyst",
+        "peds_mc_neck_mass___dermoid_cyst",
+        "peds_mc_neck_mass___enlarged_lymph_node",
+        "participant_study_id"
+        
+    ]
+    df.drop(columns=l, axis=1, inplace=True)
```
wilke0818, line 177: so drop null columns plus these ones? Why the ones above?

external_scripts/recordings_redcap/recording.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # Convert recording_id from 32 char hex to standard UUID format
```
wilke0818, line 179: is a 32 character hex, not already a UUID?

external_scripts/recordings_redcap/recording.py
```diff
@@ -0,0 +1,41 @@
+import uuid
+import pandas as pd
+import argparse
+def create_recording_redcap(recording_csv, output_csv):
```
wilke0818, line 156: either use output path and save to a specific CSV or keep the variable name but assume that they will correctly input a CSV path in

external_scripts/redcap_import.py
```diff
@@ -204,15 +204,27 @@ def clean_up(df):
...
     parser = argparse.ArgumentParser(description="List .wav files and save to CSV.")
     parser.add_argument("csv_path", help="Path to the directory to csv files")
+    parser.add_argument("uuid_json_path", help="Path to the JSON file where the record_id to uuid mapping will be stored")
     parser.add_argument("remove_fields_path", help="Path to json to remove un-needed fields")
     parser.add_argument("column_remap_path", help="Path to json remap column fields")
     parser.add_argument("instrument_mapping_path", help="insrument mapping path")
     parser.add_argument("transformer_path", help="instrument mapping path")
-    parser.add_argument("exclude_columns", help="instrument mapping path")
+    parser.add_argument("exclude_columns", help="Path to the JSON file containing the list of columns to exclude")
+    parser.add_argument("--load_uuid_map", help="Path to existing record_id to uuid mapping JSON", default=None)
```
wilke0818, line 252: what's the point of both this and uuid_json_path?

external_scripts/redcap_import.sh
wilke0818, line 1: same as above

src/b2aiprep/prepare/redcap.py
wilke0818, line 1: This does not create an order of conversation 1-4. Is there one? I assume they are ordered by the list in tdom's github? Check with them/whether these should be numbered

external_scripts/questionnaires/questionnaire.py
Evan8456: redcap imports complains if these columns are present in the csv.

external_scripts/questionnaires/questionnaire.py
Evan8456: Originally it was to merge feasibility data with the redcap but it has since changed to do more. I'll rename it.

external_scripts/questionnaires/questionnaire-example.sh
Evan8456: I'll change the files to markdown.

external_scripts/recordings_redcap/recording.py
Evan8456: It is, but Jeff was explicitly telling me how he wanted  36 char uuid (with dashes) so I explicitly convert it.

external_scripts/recordings_redcap/recording.py
```diff
@@ -28,7 +25,7 @@ def create_recording_redcap(recording_csv, output_csv):
         lambda x: str(uuid.UUID(x)) if pd.notna(x) else x
     )
 
-    df.to_csv(f"{output_csv}/recording-may13.csv", index=False)
+    df.to_csv(f"{output_csv}/recording.csv", index=False)
```
wilke0818: should also be output_path

external_scripts/recordings_redcap/recording.py
wilke0818: ok so its a validator not a hash

external_scripts/questionnaires/questionnaire-example.md
wilke0818, line 1: I was thinking more create one MD that explains how to use all these functions and why. we can never assume the users will do things right

external_scripts/recordings_redcap/recording.py
```diff
@@ -0,0 +1,198 @@
...
+    "sillySounds": "Silly Sounds",
+    "pictureDescription": "Picture Description",
+    "picture": "Identifying Pictures",
+    "days": "Days and Number naming ",
+    "months": "Days and Number naming ",
+    "sentence": "Repeating Sentences",
+    "conversation": "Conversation", 
+    "roleNaming": "Role naming tasks sounds",
+    "passage": "Reading Passages",
+    "namingAnimals": "Generative Naming Task",
```
wilke0818, line 22: do they have a consistent ordering they plan on using?

src/b2aiprep/prepare/redcap.py
```diff
@@ -292,10 +308,13 @@ def parse_audio(audio_list, dummy_audio_files=False):
             file_size = 0
         else:
             duration = get_wav_duration(file_path)
-            file_size = Path(file_path).stat().st_size
+            file_size = (Path(file_path).stat().st_size) / 1024
```
wilke0818, line 478: you sure they want 1024?

wilke0818 (approved):

external_scripts/recordings_redcap/recording.py
```diff
@@ -89,13 +90,23 @@
...
-    "roleNaming1": "Role naming tasks sounds-days",
-    "roleNaming2": "Role naming tasks sounds-months",
-    "roleNaming3": "Role naming tasks sounds-numbers",
+    "conversation(10+)1": "Conversation (10 plus)-favorite_food",
+    "conversation(10+)2": "Conversation (10 plus)-favorite_show_movie_game",
+    "conversation(10+)3": "Conversation (10 plus)-outside_of_school",
+    "conversation(10+)4": "Conversation (10 plus)-ready_for_school",
+    "conversation(6to10)1": "Conversation (6 plus)-favorite_food",
+    "conversation(6to10)2": "Conversation (6 plus)-favorite_show_movie_game",
+    "conversation(6to10)3": "Conversation (6 plus)-outside_of_school",
```
wilke0818, line 101: these are horrendous to read. is this what they want?

external_scripts/recordings_redcap/recording.py
```diff
@@ -170,12 +181,15 @@ def create_recording_redcap(recording_csv, output_csv):
...
+)   
     
+    uuid_cols = ["acoustic_task_session_id", "session_id","acoustic_task_id", "recording_acoustic_task_id", "recording_session_id"]
+    for col in uuid_cols:
+        df[col] = df[col].apply(lambda x: str(uuid.UUID(x)) if pd.notna(x) else x)
     
-    df["acoustic_task_session_id"] = df["acoustic_task_session_id"].apply(
-        lambda x: str(uuid.UUID(x)) if pd.notna(x) else x
-    )
+    # df["acoustic_task_session_id"] = df["acoustic_task_session_id"].apply(
```
wilke0818, line 192: if you're replacing with the above, delete the commented code when import finally works

external_scripts/recordings_redcap/recording.py
```diff
@@ -184,8 +198,23 @@ def create_recording_redcap(recording_csv, output_csv):
     df["recording_name"] = df["recording_name"].apply(
         lambda x: recording_names_mapping[x] if pd.notna(x) and x in recording_names_mapping else x
     )
-
-    df.to_csv(f"{output_csv}/recording-final-may21.csv", index=False)
+    
+    df["recording_input_gain"] =  ""
```
wilke0818, line 204: do we not have any sort of value for this?

external_scripts/recordings_redcap/recording.py
```diff
@@ -184,8 +198,23 @@ def create_recording_redcap(recording_csv, output_csv):
...
+    df["recording_file_share"]=   ""
+    df["recording_storage_account"] =  ""
+    
+    def acoustic_naming_fix(df):
```
wilke0818, line 208: what's the point of both this and the other mapping that you created

src/b2aiprep/prepare/redcap.py
```diff
@@ -250,16 +370,17 @@ def parse_audio(audio_list, dummy_audio_files=False):
         "conversation": [],
```
wilke0818, line 370: if this is just being run by reposchema to redcap, isn't the reposchema using the old naming? or does your external script rename them by this point?

wilke0818 (approved):
idk 🤷

src/b2aiprep/prepare/redcap.py
```diff
@@ -269,46 +404,81 @@ def parse_audio(audio_list, dummy_audio_files=False):
...
+        task_name = m.group(1) if (m := re.match(r"([a-z0-9]+(?:_[a-z0-9]+)*_\d+)", file_name, re.I)) else None
+        if task_name in task_names: # safety net; real duplicates are resolved above by _select_latest_duplicate_recordings
+            _LOGGER.warning(f"Duplicate audio task was found for {file_path}. Will be taking the first one.")
+            continue
         recording_id = re.search(r"([a-f0-9\-]{36})\.", file_name).group(1)
+        #recording_id = file_name.replace(".wav", "")
         acoustic_task = re.search(r"^(.*?)(_\d+)", file_name).group(1)
+        if acoustic_task in conversation_tasks:
+            age = get_age_from_jsonld(Path(file_path).parent)
+            if int(age) >10:
```
wilke0818: **Crash: `int(age)` on `None`.** For any conversation task, `age = get_age_from_jsonld(...)` is called just above, and that helper returns `None` whenever the session dir has no `activity_*.jsonld` or none contains an `/age` response. This line then does `int(None)` -> `TypeError`, aborting the whole `parse_audio` run. A single conversation recording (favorite_food / ready_for_school / choose_book / picture_and_doors) with no age entry is enough to trigger it. Please guard for `age is None` (skip / default cohort) before the `int()`.

src/b2aiprep/prepare/redcap.py
```diff
@@ -269,46 +404,81 @@ def parse_audio(audio_list, dummy_audio_files=False):
...
             file_size = 0
+            start_time, end_time = None, None
         else:
             duration = get_wav_duration(file_path)
-            file_size = Path(file_path).stat().st_size
+            file_size = (Path(file_path).stat().st_size) / 1024
+            start_time, end_time = get_jsonld_times(file_path)
         file_name = file_path.split("/")[-1]
+        task_name = m.group(1) if (m := re.match(r"([a-z0-9]+(?:_[a-z0-9]+)*_\d+)", file_name, re.I)) else None
+        if task_name in task_names: # safety net; real duplicates are resolved above by _select_latest_duplicate_recordings
```
wilke0818, line 482: **Silent data loss for filenames that don't match the regex.** `task_name` is `None` whenever the filename lacks a trailing `_<digits>`. After the first such file is processed, `None` is added to `task_names`, so every subsequent `None`-task file hits this `continue` and is dropped (only a warning). Note `_select_latest_duplicate_recordings` deliberately *keeps* all `task_name is None` paths, so this safety net silently undoes that. Consider skipping the check when `task_name is None`.

src/b2aiprep/prepare/redcap.py
```diff
@@ -680,10 +861,14 @@ def _convert_reproschema_to_redcap(
                 "redcap_repeat_instrument": "Session",
                 "redcap_repeat_instance": 1,
                 "session_id": audio_session_ids.get(subject_id, None),
+                "session_record_id": subject_id,
                 "session_status": "Completed",
                 "session_is_control_participant": "No",
-                "session_duration": session_durations[subject_id],
+                "session_duration": session_durations[subject_id]["session_duration"],
```
wilke0818, line 924: **Potential `KeyError`.** `session_durations` is keyed by `record_id` from the *questionnaire* subject folders, but this dict is built while iterating `audio_sub_folders` (`subject_id` from audio files). A subject with audio but no questionnaire session (or a name-normalization mismatch) is absent from `session_durations`, so `session_durations[subject_id]["session_duration"]` throws. Suggest `.get(subject_id, {"session_duration": np.nan, ...})`, mirroring how `audio_session_ids.get(...)` is already handled two lines up.

src/b2aiprep/prepare/redcap.py
```diff
@@ -269,46 +404,81 @@ def parse_audio(audio_list, dummy_audio_files=False):
...
         session = Path(file_path).parent.name
         if dummy_audio_files:
             duration = 0
             file_size = 0
+            start_time, end_time = None, None
         else:
             duration = get_wav_duration(file_path)
-            file_size = Path(file_path).stat().st_size
+            file_size = (Path(file_path).stat().st_size) / 1024
+            start_time, end_time = get_jsonld_times(file_path)
```
wilke0818, line 479: **Redundant repeated I/O.** `get_jsonld_times` opens and `json.load`s every `activity_*.jsonld` in the dir, and is called both here in the main loop and inside `_select_latest_duplicate_recordings` for each duplicate — up to twice per call (exact-match pass + substring fallback). For N recordings and M activity files this is O(N*M) file opens. Consider caching parsed JSON-LD per session dir.

external_scripts/redcap_import.py
```diff
@@ -195,24 +195,61 @@ def clean_up(df):
...
+    def gen_uuid(id):
+        if pd.isna(id):
+            return id
+        if id not in record_id_2_uuid_tracker:
+            record_id_2_uuid_tracker[id] = str(uuid.uuid4())
+        return record_id_2_uuid_tracker[id]
+
+    df["participant_study_id"] = df["record_id"]
+    df['record_id'] = df['record_id'].apply(gen_uuid)
+    df['session_record_id'] = df['session_record_id'].apply(gen_uuid)
```
wilke0818, line 240: **Assumes `session_record_id` column exists.** `df['session_record_id'] = df['session_record_id'].apply(gen_uuid)` raises `KeyError` if the input CSV lacks that column. Unlike `record_id`, it's only present when the CSV came from the updated `redcap.py`; older/partial exports will crash here.

external_scripts/questionnaires/questionnaire.py
```diff
@@ -0,0 +1,148 @@
...
+                    "ef_primary_language": "1",
+                    "ef_enrollment": "Enroll",
+                    "ef_is_control_participant": "no",
+                    "ef_enrollment_institution": "sickkids",
+                    "enrollment_form_complete": "2"
+                    
+                }
+            ]
+        )
+        df = pd.concat([feas,df], ignore_index=True)
```
wilke0818, line 103: **O(n^2) concat + UUID coercion crash.** `pd.concat([feas, df], ignore_index=True)` inside the per-subject loop rebuilds the entire frame every iteration (quadratic in #subjects); consider collecting the `feas` rows in a list and concatenating once. Separately, the `str(uuid.UUID(str(x)))` applied to every `*session_id*` column (line ~127) raises `ValueError` on non-canonical values — e.g. the trailing-dash session ids like `...-566-` emitted by `redcap.py` are not valid UUIDs.

external_scripts/recordings_redcap/recording.py
```diff
@@ -0,0 +1,227 @@
...
+        
+    ]
+    df.drop(columns=l, axis=1, inplace=True)
+
+    # Convert recording_id from 32 char hex to standard UUID format
+    # df["recording_id"] = df["recording_id"].apply(
+    #     lambda x: str(uuid.UUID(x)) if pd.notna(x) else x
+    # )
+    
+    df["recording_id"] = df["recording_id"].apply(
```
wilke0818, line 184: **Fixed column list + stray `-` prefix.** The `uuid_cols` loop `df[col].apply(...)` `KeyError`s if any of those columns is absent. Also `re.sub(r"([a-f0-9]{32})", lambda m: f"-{uuid.UUID(...)}", x)` prepends a literal `-`, producing `recording_id` values like `-d6d7411f-...`, which won't match the 36-char UUIDs used elsewhere. Was the leading `-` intended?

wilke0818 (commented):
Automated code review (high effort). 7 findings below, ranked by severity. Notable non-findings — the `acoustic_task_id` -> session change, `Pediatrics`->`Pediatric`, and `file_size / 1024` (bytes->KB) all look intentional and match the updated tests.

src/b2aiprep/prepare/redcap.py
Evan8456: it uses the new fields since i remap during the generation process.

external_scripts/recordings_redcap/recording.py
Evan8456: extra remappings that Jeff wanted

external_scripts/recordings_redcap/recording.py
Evan8456: Nope reporshema doesn't have this field.

external_scripts/recordings_redcap/recording.py
Evan8456: Yeah, I fought hard, but they were adamant that it matched the app.

Evan8456 (commented):
looks reasonable. We should test the imports and bids once data is fully uploaded.

Evan8456 (commented):
so this in theory should fail? the raw data, conversation is favorite_food_6_to_10, favorite_movie_6_to_10 etc. but we later remap it to conversation. so conversation will never increment

wilke0818:
I ran the findings against the actual 806 dataset rather than reasoning in the abstract, which changes what's actually blocking. One naming mismatch fires on this data; the two crash paths don't; and there are two design questions worth settling before the import round-trips into the shared export path.

Context that shaped the review: inside `parse_audio`, `is_import` gates only the remap (`redcap.py:547-548`) and one `race` branch (`:186`). The age→band logic, session-scoped ids, consolidation, and numbering all run on both the import and direct paths.

### Fires on this data — worth fixing here

**Singular/plural `pictures_and_doors`** (`redcap.py:387`, `recording_name_remap.json`). The dataset has **135** `pictures_and_doors_*` wavs and **0** singular `picture_and_doors`, but `conversation_tasks` uses the singular. As a result:

- the membership check misses, so these never get rewritten to `conversation(2to4)`;
- `"picture"` is a substring of `"pictures_and_doors"`, so `protocol_order` buckets them into Identifying-Pictures;
- the remap key `pictures_and_doors` has no `-N`, but `parse_audio` emits `pictures_and_doors-1`, so the remap misses too.

Net: 135 book-task recordings come out mislabeled. Fix is a stem correction + a `-N`-aware remap key.

### Design questions

- **Session-scoped ids** (`redcap.py:506`, `:555`): `acoustic_task_id`/`recording_acoustic_task_id` are now just `{session}`, so every task/recording in a session shares one id. That may be right for a REDCap import (linkage is record_id + instance), but the same values feed the recordings→tasks join at `redcap.py:1070-1072` — if the direct (`is_import=False`) path still uses that join, it becomes many-to-many within a session. Is the session-scoped id intended, and is the direct path expected to use that join? The test locks it in, so worth confirming it's deliberate.
- **Raw vs canonical names on import**: the remap is `is_import`-gated, so the import writes raw names into REDCap. These are high-count in the data — `naming_animals` (511), `naming_food` (503), `pictures_and_doors` (135), `123s` (133), `months` (86), `days` (85), `abcs` (56). Downstream resolution is case-insensitive substring on the token, not case-sensitive. Should the import write raw names and rely on REDCap/the dictionary to canonicalize before export, or write canonical names directly? This determines whether the imported data re-exports into the same shape as the existing REDCap exports.

### Latent — verified they don't fire on this dataset, but unguarded

- `int(age)` on `None` (`redcap.py:489`): all 802/802 audio sessions carry an `/age` Response (value int-able, e.g. `'9'`), so no crash here — but a subject missing age aborts the whole parse.
- `re.search(...).group(1)` (`redcap.py:485`): 0 of 75,937 wavs lack a `_<digit>` stem, so it doesn't trigger — but the `.group(1)` is unguarded.
- `redcap_import.py` (`:198`, `:240`): fine for well-formed input; KeyErrors if `age`, `session_record_id`, or the `*_complete` columns are absent.
- jsonld helpers (`redcap.py:269`, `:477`): re-glob/re-read every `activity_*.jsonld` per call (twice for duplicates), and `json.load` is unguarded — one malformed file aborts the parse. Parse once into a filename→times map + try/except.

Out of scope for this PR: the description-matcher change and the `audio_task_descriptions.json` corrections (CAPE-V/loudness/random) are on a separate branch.

The `pictures_and_doors` mismatch is the only item I'd treat as blocking on this data; the two design questions are the calls that decide whether the import re-exports cleanly. The rest is hardening for other datasets.

wilke0818:
Follow-up on the naming — I compared what this branch emits against the existing REDCap-export peds BIDS release (07_07) to check the import→export round-trip. It mostly reconciles.

### Confirmed working
- `pictures_and_doors` is now consolidated and named correctly (e.g. `Conversation-(4-to-6)-book`), and the duplicate/retake handling (`_select_latest_duplicate_recordings`) picks a single take and warns.
- Names that match the export exactly: `Abcs-and-123s-abcs/123s`, `Conversation-(2-to-4)/(4-to-6)-book`, `Conversation-(6-plus)-*`, `Generative-Naming-Task-animals/food`, `Identifying-Pictures-N`, `Long/Noisy/Silly-Sounds-N`, `Reading-Passage-N`, `Repeating-Sentences-N`, every `Repeating-Words-<word>` form, and `Role-naming-tasks-sounds-days/months/numbers`.

### Two mismatches vs. the export
1. **`Conversation-(10-plus)-*`** — for the 10+ band this branch emits `Conversation-(10-plus)-{favorite-food,favorite-show-movie-game,outside-of-school,ready-for-school}`, but the export (and the redcap docs) only define `Conversation-(6-plus)-*`; there is no "10 plus" conversation task. The 10+ band should map to `6-plus`, unless a 10-plus split is intended.
2. **`days` / `months` leak** — `parse_audio` emits `days-1` / `months-1`, but `recording_name_remap.json` keys them as bare `days` / `months`, so the remap misses and the raw names pass through. The export has `Days-and-Number-naming-days` / `-months`. The remap values are already correct; the keys just need to be `days-1` / `months-1` (the same `-N` form already applied to `abcs-1`/`123s-1`).

### Minor / to confirm
- `Picture Description` is emitted with a space vs. the export's `Picture-Description`; likely normalizes to a dash downstream, but inconsistent as a `recording_name`.
- The export has `Days-and-Number-naming-numbers`, which this branch doesn't emit — but that appears to be because the input has only `days`/`months` recordings for that task (no `numbers` recording present), so it's an input gap rather than a code issue. Worth confirming.

Net: after the `10-plus` band mapping and the `days`/`months` remap keys, the emitted names line up with the export for every task family present in the input.

PR #308, opened 2026-05-20
Updates for internal releasing 

wilke0818:
- Adds a command for creating subject splits, used for running feature generation more efficiently (could've added to external scripts but decided it doesn't hurt to have here)
- Renamed `id_remap` to match verb-based command naming
- Added remapping IDs to RELEASE.md


Evan8456 (approved):
Looks good.

PR #312, opened 2026-06-03
Normalize task_name slug in audio_quality_metrics.tsv

wilke0818:
quality_control.py extracted task_name from the (pre-deidentification) BIDS audio path verbatim, leaving capitalized/parenthesized recording names like 'Cape-V-sentences-(v2)-1'. Every other task-bearing output ends up slugged (deidentification renames audio/feature files via sanitize_task_entity_in_bids_stem), so audio_quality_metrics.tsv was the lone file with raw names, which broke exact task-name joins downstream. Apply normalize_task_label after extraction so the metrics use the same slug as the rest of the dataset.


Evan8456 (approved):
lgtm

PR #317, opened 2026-08-04
Map redcap 4.9.1 fields into bids_field_organization.csv

wilke0818:
Stacked on #316. The base of this PR is `redcap2rs-4.9.1-20260721162303` — #316's own branch — not `main`. **Merge this PR first**: it lands these commits on #316's branch, and #316 then carries both changes into `main` as one unit. Merging #316 first would strand this PR against a branch that no longer feeds anywhere.

## Why

#316 moves the vendored reproschema from redcap 4.7.0 to 4.9.1, but `bids_field_organization.csv` was last synced at 4.7.0 (`9c017c6`). `BIDSDataset._construct_phenotype_from_reproschema` iterates the **CSV**, not the reproschema, so the 262 data elements the bump adds have no row and would be absent from `phenotype/` entirely — with nothing but an INFO log line (`excluded_in_df_not_in_reorg`, `dataset.py:1015`) to show it. Omission is fail-safe rather than a leak, so this is silent data loss, not a privacy problem.

No CSV row lost its upstream element, so the change is additive apart from one retirement.

## What this adds

385 rows (2918 active / 456 withheld across the whole file):

| Source | Rows | Destination |
|---|---|---|
| `d_neuro_ataxia` | 117 active (60 checkbox expansions) | `phenotype/diagnosis/ataxia` |
| `d_neuro_essential_tremor` | 114 active (63 expansions) | `phenotype/diagnosis/essential_tremor` |
| `subjectparticipant_eligible_studies` | 2 | `eligibility` → `self_reported_neuro_ataxia`, `self_reported_essential_tremor` |
| `enrollment_form` | 2, `delete=YES` | duplicates eligibility, matching existing `ef_parkinsons` |
| `recording` | 10 | `recording`/`task` — device and browser telemetry added at 4.8.0 |
| `prolific_information` | 3 (1 active) | `is_prolific` → `recruited_via_prolific`; `prolific_id` and `show_prolific_study_id` withheld |
| 6 new econsent activities | 145, `delete=YES` | consent text, names, dates, signatures — never released |

Conventions followed from the existing file: a `VARIABLE` row for each checkbox base plus one `CHECKBOX_OPTION` row per option (all 205 existing bases do this); `_gsd` spelled out as `_gold_standard_diagnosis` in `column_name`, per `diagnosis_als_gsd_confirmation_method___*`; `descriptive` REDCap fields as `delete=YES` since they never appear as data columns.

### Retired

`diagnosis_healthy_developmentally_typical_children_ci` → `delete=YES`. Its activity was dropped from `b2ai-redcap2rs` at redcap 4.8.0 and is no longer in the protocol `ui.order`, so the row can no longer resolve in `element_to_schema` (`dataset.py:872`). It doesn't crash today only because the field is absent from every 4.x data dictionary (checked v4.0.0 → v4.9.1) and so trips the `col_norm not in df_cols` guard first. No data is lost.

## Withheld, and one set to revisit

`recording_device_info`, `recording_screen_resolution`, `recording_window_resolution` and `recording_platform_architecture` are **released** (`delete=NO`). Individually innocuous, but together they form a browser fingerprint, so whether they stay released depends on what the app actually writes into them — revisit once real values are available.

`prolific_id` is **withheld**. A Prolific ID is a stable identifier assigned by an external platform and reusable across studies, so it is directly re-identifying — not a judgement to defer. `is_prolific` stays released: recruitment source is a category, not an identifier.

Note that `delete=YES` *is* the deidentification mechanism — see the comment at `dataset.py:1397-1401` ("The deidentification is now implicitly applied by the use of bids_field_reorganization.csv to select only desired columns") — so flipping any of these back is the whole control.

`show_prolific_study_id` is **withheld**. Nothing in `bridge2ai-redcap` defines it: no field note, no branching logic anywhere in the project, no action tag, and its Spanish MLM entry is untranslated. There's no basis for judging whether its values are releasable. Whoever owns the Prolific recruitment path in the web app, or the WCM admin behind `bridge2ai-redcap@9797f48`, can settle it.

## No `RepeatInstrument` entries

None of the three new forms is a repeating instrument in the REDCap project definition — the XML declares exactly 31, and they are `session`, `acoustic_task`, `recording` and the `q_*`/`pediatric_q_*` questionnaires. REDCap therefore leaves `redcap_repeat_instrument` blank on their rows, `RedCapDataset._load_redcap_csv` fills the blanks with `"Participant"` (`redcap.py:492`), and that maps to `subject_participant_basic_information_schema`. A `"D Neuro Ataxia"` or `"Prolific Information"` label can never appear, so an entry would add nothing but an `instrument_columns/*.json` for `Instrument.__init__` to load eagerly.

Row separation is done by `dropna(how="all", …)` at `dataset.py:987`, which is also why `parkinsons_disease.tsv` works today despite `parkinsons_disease_schema` never appearing in the mapped column either. Confirmed by running `_construct_phenotype_from_reproschema` against a synthetic export: `diagnosis/ataxia.tsv` and `diagnosis/essential_tremor.tsv` are both produced, the `_gsd` rename lands, checkbox columns binarize to `1`, and participants land in the correct file.

## Tests

`tests/test_bids_field_organization.py` (7 tests) guards coverage in both directions plus the invariants the phenotype writer relies on: unique `(schema_name, column_name)` per schema (else `dataset.py:972` raises), one `group` per schema, every `___option` row having a base row, no empty descriptions. Two of the seven fail on this branch's parent, so they would have caught this bump.

## Verification

- `159 passed` across the suite; the one `ERROR` (`test_prepare.py::test_extract_features_timing`) is a missing `pytest-benchmark` fixture and is identical on the #316 baseline.
- `test_cli.py` green, including `test_redcap2bids_cli`, which drives the CSV path end to end.

## Follow-ups, deliberately not in this PR

- **`pipeline_trigger.yml:52-56` syncs with `cp -r` and never deletes.** Every upstream activity removal leaves residue; `activities/pediatric_d_healthy_developmentallytypical_childre/` is the current one, still present though unreachable.
- **`update-bids-template` is dead code.** Its output directory `src/b2aiprep/template/phenotype/` holds nothing but `.gitkeep` (emptied by `61674ff`), nothing consumes it, and `dataset.py:279` copies the empty directory into every BIDS output. `update.py:266` reads a `renamed_column` key that isn't in the CSV, so `--reorganize` raises — worth deleting rather than fixing, since `6dd24ae` superseded the whole template pipeline.
- **Legacy `schema_name_source` values.** They are REDCap *form names* minus the category prefix, not reproschema activity ids — 61 of 68 match that pattern, and `alzheimers_disease_mild_cognitive_impairme_schema` carries REDCap's form-name truncation. Only 4 match `RepeatInstrument.schema_name`, so the repeat-instrument filter is inert for every non-repeating form. Correcting all 55 buys accurate documentation and zero behaviour change; the better fix is dropping the misleading `schema_name_pretty` values on the 45 non-repeating entries, or the `have_all_schema_names` branch itself, so nobody reads that code and assumes filtering happens.


wilke0818:
@alistairewj @satra currently setting `recording_device_info`, `recording_screen_resolution`, `recording_window_resolution` and `recording_platform_architecture` to release as Alex said they were approved by ethics, however, they do seem like they might be PII/possibly identifying when put together

satra:
@wilke0818 - if ethics approved it, we go with that SOP. perhaps check with hortense. those by themselves are not PII/PHI in general. IP address would be.

Evan8456 (approved):
lgtm

PR #321, opened 2026-08-12
Task registry: generated registry-first resolver + doc-faithful curation (Phase 2)

wilke0818:
## Summary
Phase 2 of the audio-task metadata redesign. Replaces the fragile flat `audio_task_descriptions.json` substring lookup with a **generated, vendored task registry** that mirrors the `bridge2ai-redcap` hierarchy, plus a **registry-first resolver** (with flat-file fallback so nothing regresses), and corrects instructions/stimuli 1-by-1 against the redcap docs.

Stacked on top of **Phase 1** (`fix/audio-task-descriptions-matcher`), which targets `main`.

## What changed
- **Registry generator** (`scripts/build_task_registry.py`, dev/CI only) + vendored `resources/task_registry/` (registry.json + stimulus banks); `--check` drift guard; packaged via `pyproject`.
- **Registry-first resolver** in `fhir_utils.py` (exact → longest normalized-substring alias match; per-recording instructions; three-tier fallback to the flat file).
- **Sidecar model**: the legacy `prompts` array is replaced by scalar **`stimulus_text`** + **`speech_type`** (read/recall/elicited/non-lexical) + **`stimulus_source`** + **`stimulus_asset`**.
- **Curated, doc-faithful instructions** (authoritative over harvested text): adult v1 (retired) story-recall (+ its recall passage), maximum-phonation-time, diadochokinesis, respiration-and-cough, breath-sounds, free-speech; the respiration v2 name/instruction swap; the grouped pediatric tasks; pediatric picture-description + identifying-pictures.
- **Image `stimulus_asset`**: commit-pinned raw GitHub URLs for every image-stimulus task (reading-passage, repeating-sentences, noisy/silly/long-sounds, identifying-pictures, picture-description).
- **Cohort-aware routing**: `picture-description` exists in both populations; the resolver routes by the acoustic task's cohort (peds vs adult).
- **Non-lexical stimulus**: static non-lexical tasks now emit `stimulus_text=""` instead of leaking the instruction.
- **Build-time validation/warnings**: version-aware matching, recording-numbering guard (impossible indices), unknown-task warnings, and within-session recording-name collision detection.
- **Bundle**: per-recording metadata table emitted as `metadata.tsv` (was parquet) with an updated data dictionary; the c2m2 bundler updated to read it.

## Schema note (downstream)
Recording-metadata sidecars now carry scalar `stimulus_text` + `speech_type` (+ `stimulus_source`/`stimulus_asset`) instead of a `prompts` array; the bundled metadata table is now TSV. No in-repo reader of the old fields remains.

## Testing
- Unit/integration: `tests/test_task_prompts.py`, `tests/test_task_registry.py`, `tests/test_resources.py` (+ full BIDS pipeline test) — all green; generator `--check` clean; wheel builds with the vendored registry.
- **Real-data verification**: a task-by-task reconciliation of every registry task against the redcap docs passed; all four BIDS datasets (two adult, pediatric, full-reproschema) were regenerated on this branch with **zero unmatched tasks** across ~165k recording sidecars, and the corrections confirmed present.

## Out of scope / follow-ups
- The optional "compat regen" collapse of the flat file is intentionally held.
- A few cosmetic doc-wording nits left as-is (e.g. cape-v v1 "sentences" vs doc "passage").
- Two upstream (redcap data, not code) items tracked separately: a set of malformed recording names, and a batch of empty-at-source audio files.


Evan8456:
Code Review: fix/audio-task-descriptions-matcher (vs main)
=============================================================
Scope: committed diff only (untracked local files excluded).


### Correctness

**1. `src/b2aiprep/prepare/fhir_utils.py:333`**
Alias-following speech_type fallback classifies using the pre-alias task key instead of the resolved target, so future aliases whose name doesn't share their target's prefix get silently misclassified.

> Adding an alias like `"cinderella-retell": {"alias_of": "cinderella-story"}` (a recall-family target) causes `_classify_speech_type` to be called with `"cinderella-retell"`, which matches no prefix and silently defaults to `"elicited"` instead of `"recall"`. Latent today because the existing aliases happen not to trigger it — worth fixing so `best_task` is reassigned to the resolved target name before classification.

---

**2. `src/b2aiprep/prepare/resources/audio_task_descriptions.json:4405`**
Static fallback `prompts` for random-item-generation now contain two mutually-exclusive prompt variants that get concatenated into one incoherent `stimulus_text` when no questionnaire row matches.

> When `_stimulus_text_from_questionnaire` returns `None` (e.g. the NEURO_RANDOM_ITEM_GENERATION instrument is missing for that cohort), `stimulus_text` becomes both the letters/numbers prompt AND the category prompt joined together, even though the participant only performed one. This is a real regression from the old empty-list fallback.

---

**3. `src/b2aiprep/prepare/dataset.py:227`**
`questionnaire_lookup` is built with `pd.notna(task_id)` instead of the `_is_present()` helper introduced in this same diff, so the literal string `"nan"` is treated as a present join key.

> `redcap.py`'s `_clean_reproschema_df` calls `.astype(str)` on `acoustic_task_id` columns, turning real NaNs into the string `"nan"` before it's written to the intermediate CSV, which is then read back without `"nan"` parsed as missing. `pd.notna("nan")` is `True`, so rows with missing task ids get inserted into `questionnaire_lookup` keyed by `(instrument, "nan")`, silently colliding. Currently inert only because the lookup side already rejects `"nan"` via `_is_present`, but the two NaN-detection idioms are inconsistent within the same PR — worth using `_is_present()` here too for consistency.

---

**4. `src/b2aiprep/prepare/dataset.py:228`**
`questionnaire_lookup` silently overwrites on duplicate `(instrument, task_id)` keys, unlike the adjacent `audio_files_by_recording` construction which warns on duplicates.

> A duplicate submission/retry/data-entry error producing two rows for the same `acoustic_task_id` causes the earlier row's stimulus data to be silently dropped with no log line, inconsistent with the nearby `audio_files_by_recording` loop (dataset.py:260-266) which explicitly logs "Multiple audio files found for recording UUID..." in the analogous case.


### Documentation

**5. `src/b2aiprep/prepare/resources/metadata.json:36`**
The shipped data dictionary still documents the removed `Prompts` field and wasn't updated for the new `stimulus_text`/`speech_type` columns this diff adds.

> `commands.py` copies `resources/metadata.json` verbatim into every output dataset's `metadata/` folder alongside `metadata.parquet`, which now includes `stimulus_text` and `speech_type` columns. Downstream users see a stale `Prompts` entry and no documentation for the two new columns.


### Efficiency

**6. `src/b2aiprep/prepare/fhir_utils.py:292`**
Per-recording task-description resolution now does a full substring scan over the ~300-entry description dict per call instead of breaking on first match as before.

> For datasets with many thousands of recordings, each conversion call rescans the whole static dict with no caching or pre-sorting, adding avoidable CPU work to a per-recording hot path.


### Simplification / cleanup

**7. `src/b2aiprep/prepare/fhir_utils.py:330`**
`static_prompts[0] if len(static_prompts) == 1 else " ".join(static_prompts)` is redundant — `" ".join(static_prompts)` already produces the same result for 0, 1, or N elements.

> No functional bug; the `len == 1` special case is unnecessary complexity since `" ".join(["x"]) == "x"` already.

---

**8. `src/b2aiprep/prepare/fhir_utils.py:107`**
Vocab/random/stroop stimulus resolution is three hardcoded if-branches on string prefixes rather than extending the existing data-driven `prompt_ref` registry used elsewhere in this same diff.

> Adding a fourth questionnaire-backed task requires editing Python in two places (dataset.py's join tuple and fhir_utils.py's if-chain) instead of adding a declarative entry to `audio_task_descriptions.json`, unlike `prompt_ref` which already generalizes bank-based lookups.

wilke0818:
Thanks for the review. Note the diff you reviewed (`fix/audio-task-descriptions-matcher` vs `main`) is actually the **#322** changeset, so fixes are split across the two stacked PRs by where the code lives — findings that touch the still-relevant flat/dataset code went on phase-1 (#322, `be78ccf`), and findings in the resolver rewrite went on phase-2 (#321, `c013bfa`; merged up in `f77a61d`).

| # | Finding | Resolution |
|---|---|---|
| 1 | speech_type classified on pre-alias key | Fixed on #321 (`_flat_bids_fields`): `best_task` is reassigned to the resolved alias target before classification/join. |
| 2 | random-item static `prompts` concatenated into one incoherent stimulus | Fixed on #322: dropped the two mutually-exclusive variants (`prompts: []` + explicit `speech_type: elicited`); the per-participant category comes from the questionnaire join. Added a regression test asserting empty `stimulus_text` when no row matches. |
| 3 | `pd.notna` vs `_is_present` for the string `"nan"` | Fixed on #322: the join now uses `_is_present(task_id)`. |
| 4 | silent overwrite on duplicate `(instrument, task_id)` | Fixed on #322: warns and keeps the first, matching the adjacent `audio_files_by_recording` handling. |
| 5 | `metadata.json` still documents removed `Prompts` | Fixed on both: the data dictionary now documents `stimulus_text`/`speech_type`/`stimulus_source`/`stimulus_asset` (already on #321; backported identically to #322 so it's correct standalone). |
| 6 | full substring scan, no early break | No change needed: the loop already `break`s on an exact match; the remaining pass is the intentional longest-substring match, which can't short-circuit. |
| 7 | redundant `len==1` special case | Fixed on #321: replaced with `" ".join(static_prompts)`. |
| 8 | hardcoded vocab/random/stroop if-chain | Deferred: making questionnaire-join a declarative `prompt_ref` type is a non-functional refactor; tracking separately rather than in this PR. |

All green: 35 tests on phase-2, 10 on phase-1; both PRs `mergeable`.

Evan8456:
# Code Review: PR #321 — Task Registry (Phase 2)

**Repo:** sensein/b2aiprep
**PR:** [#321 — Task registry: generated registry-first resolver + doc-faithful curation (Phase 2)](https://github.com/sensein/b2aiprep/pull/321)
**Branch:** `feat/task-registry` → `main`
**Reviewed:** committed diff only (untracked local files excluded)
**Diff size:** 20 files changed, ~4,985 insertions / 49 deletions (bulk of it new static JSON banks + the vendored `registry.json`)

## Summary

Overall the branch is in good shape. Every substantive bug found during review had already been fixed in follow-up commits (`c013bfa`, `be78ccf`, `01f7023` — all titled "Address review: ...") pushed after my initial pass. Two minor, non-blocking items remain. No blocking issues.

## What changed

- `src/b2aiprep/prepare/fhir_utils.py` (+658/-…): new registry-first task resolution. Resolves a recording's `task_name` against a vendored `task_registry/registry.json` first (authoritative for instructions, `speech_type`, and stimulus `prompt_ref`), falling back to the legacy flat `audio_task_descriptions.json` for anything the registry can't produce (e.g. static passage/recall text). Replaces the old list-valued `prompts` field with a scalar `stimulus_text` + new `speech_type`/`stimulus_source`/`stimulus_asset` fields.
- `src/b2aiprep/prepare/dataset.py` (+70/-…): builds a `questionnaire_lookup` (keyed by `(instrument, acoustic_task_id)`) for the three tasks whose per-participant stimulus lives in a linked REDCap questionnaire (vocab words, random-item category, stroop colors) rather than a static description. Also adds recording-name collision detection within a session (warns instead of silently overwriting a sidecar/audio file).
- `src/b2aiprep/commands.py`: metadata output switched from `metadata.parquet` to `metadata.tsv` (flat scalar fields no longer need parquet's nested-type support).
- `external_scripts/c2m2/bundle_to_c2m2.py`: updated to read either `metadata.tsv` or legacy `metadata.parquet`.
- `scripts/build_task_registry.py` (new, 714 lines): offline generator that builds `registry.json` from the bridge2ai-redcap doc hierarchy. Build-time only, not on the runtime hot path.
- New static resources: `task_registry/registry.json`, several stimulus "bank" JSON files (harvard sentences, productive vocabulary, CAPE-V sentences, etc.), and `task_instructions_curated.json`.
- Tests: `test_task_registry.py`, `test_task_prompts.py`, `test_resources.py` additions covering resolution and registry integrity.

## Issues found (all resolved by the time of final review)

These were real bugs on my first pass through the pre-fix version of the branch. All four were fixed in the author's subsequent "Address review" commits — confirmed by diffing the current `origin/feat/task-registry` against my findings.

| # | File | Issue | Fix confirmed |
|---|------|-------|----------------|
| 1 | `fhir_utils.py` — `_flat_bids_fields` | Alias-following `speech_type` classification used the **pre-alias** task key instead of the resolved target, so an alias whose name doesn't share its target's family prefix would be silently misclassified (e.g. defaulting to `elicited` instead of `recall`). | `best_task` is now reassigned to the alias target before classification. |
| 2 | `dataset.py` — `questionnaire_lookup` construction | Used `pd.notna(task_id)` instead of the new `"nan"`-string-aware `_is_present()` helper. Since an upstream `.astype(str)` step can turn real NaNs into the literal string `"nan"`, `pd.notna("nan")` is `True` and such rows were incorrectly treated as present. | Now uses `_is_present(task_id)`. |
| 3 | `dataset.py` — `questionnaire_lookup` construction | Duplicate `(instrument, task_id)` keys were silently overwritten with no log line, unlike the adjacent `audio_files_by_recording` construction which warns on duplicates. | Now logs a warning ("Multiple {instrument} questionnaire rows for acoustic_task_id ...; keeping the first, ignoring duplicate.") and keeps the first row. |
| 4 | `resources/metadata.json` | Shipped data dictionary still documented the removed `Prompts` field and had no entries for the new `stimulus_text`/`speech_type` columns. | Data dictionary updated with `instructions`, `stimulus_text`, `speech_type`, `stimulus_source`, and `stimulus_asset` entries. |
| 5 | `fhir_utils.py` — `_flat_bids_fields` | `static_prompts[0] if len(static_prompts) == 1 else " ".join(static_prompts)` was a redundant special case — `" ".join(static_prompts)` already produces the identical result for 0, 1, or N elements. | Simplified to `" ".join(static_prompts)`. |
| 6 | `bundle_to_c2m2.py` docstring | Stale comment referenced `metadata.parquet` only, not reflecting the new `.tsv` output. | Comment updated to mention both. |

## Remaining minor items (non-blocking)

1. **`fhir_utils.py` — `_resolve_task_registry` substring-match loop.** Iterates `_alias_items()` (sorted longest-alias-first) looking for a substring match but doesn't break early once no further match can improve on the current best. Harmless in practice — the list is `lru_cache`d and short — but slightly wasteful per call. Not worth blocking on.
2. **Not deep-reviewed:** `scripts/build_task_registry.py` (714-line new generator script) and the large static JSON banks (`registry.json` at 1,555 lines, `harvard_sentences_bank.json` at 869 lines, etc.). These are build-time artifacts / data files rather than runtime logic, so I only skimmed structure. Flag if you want a line-by-line pass on these too.

## Recommendation

**Approve.** The registry-first resolution design is sound (registry authoritative where it has curated data, flat-file fallback preserves existing behavior elsewhere), and the review-cycle fixes already landed address every correctness issue I found.

PR #326, opened 2026-08-17
Language-aware task metadata: record session language + Spanish (es-419) instructions & stimulus

wilke0818:
## Summary

Makes the audio-task metadata resolver **language-aware**. Every recording sidecar now records the session's administration language, and Spanish (`es-419`) sessions receive Spanish instructions and Spanish stimulus (WER reference) text instead of inheriting English. Fixes the ~316 adult recordings that were carrying English references for Spanish speech, and makes non-English recordings filterable.

Stacks conceptually on the merged task-registry work (#321/#322); branched off `main`.

## What changed

- **`language` field on every sidecar** (`en` / `es-419` / `fr-CA`), from the RedCap `selected_language` radio. Mapped from text labels, integer codes (`1/2/3`), and BCP-47 forms, so a *coded* export never silently collapses Spanish/French to English; unrecognized values warn and default to `en`. Documented in `metadata.json`.
- **es-419 stimulus** for read/recall tasks that have a Spanish source: Harvard + CAPE-V banks (`*_es_419.json`), and family-keyed static text for Caterpillar + Story Recall. Resolution falls back to the English bank when no `es-419` variant exists, and **warns** for read/recall so an English reference is never silently shipped tagged as another language.
- **Spanish instructions** for all 14 current-protocol families with an es-419 description (`task_instructions_es_419.json`).
- **`scripts/harvest_es419_banks.py`** parses the committed bridge2ai-redcap `es-419` docs into the banks (no hand-transcription of WER references); `--check` drift-guards them.
- **Random Item Generation** instruction is now category-driven and variant-aware: a semantic category (Category_2 only) → the non-repeatable instruction; Numbers/Letters (in both category lists, undeterminable) → the general instruction; the drawn category is surfaced in all cases. The **time limit was removed** (`random_duration` shows it was never an enforced cap).

## Safety / scope

- **English `stimulus_text` is unchanged** — every es-419 code path is gated on `language != "en"`. Two intended English-side changes ship in this PR: the added `language` field, and the Random-Item *instruction* (category-driven; the time limit was removed per request, which makes v1/v2 instructions identical). English `stimulus_text` and all other instructions are unchanged.
- **Pediatrics unaffected**: peds carries the same `selected_language` (all English) → `en`, so all Spanish machinery is gated off; a missing value also defaults to `en`.
- The questionnaire join (vocab/random/stroop) is language-agnostic (verified); no Spanish session recorded those tasks in 07_01, but a future one resolves correctly.

## Verification

- 43 tests pass (`test_task_prompts` / `test_task_registry` / `test_resources`); `harvest --check` drift-clean.
- es-419 stimulus verified against real recording names; export and built phenotype counts agree.

Design notes and open follow-ups (Free-Speech cue still English; fold the harvest into `build_task_registry.py`; the es-419/English time-limit source discrepancy) are in `docs/design/language-aware-tasks.md`.


wilke0818:
**Addressed the high-effort code review** (all pushed in `402738e`):

- **Warn on non-English read/recall with no language reference** (static-inline path) — symmetric with the stimulus-bank path. Rainbow/Cinderella in a Spanish session no longer silently ship the English text tagged `es-419`.
- **Free-speech Spanish cue scoped to the Free Speech family** — no other `(v2)-N` static-inline task can be handed the free-speech question.
- **Lowercase the language slug** in bank/resource lookups — a future region-cased code (e.g. `fr-CA`) matches lowercase bank files instead of silently falling back to English.
- **`_random_item_instruction` uses `.get()`** — a malformed language entry degrades gracefully instead of `KeyError`-crashing the build.
- **Kept** (deliberate): blank `selected_language` → `en` silently — sensible default, always populated in the real data, and a per-recording warning would be noise. Also confirmed `selected_language_2` is never loaded, so the field-choice concern is moot.

**Revalidated against the real build** (`bids_07_01_lang`), not just unit tests: 16/16 Spanish participants and 657/657 sidecars tagged `es-419` (exact match to redcap); CAPE-V (v1-labelled, via version-fallback), Harvard, Caterpillar, and Story-recall all carry Spanish `stimulus_text`; only Rainbow (retired, no es source) falls back to English (now warned); English participants unchanged. The v1-labelled-CAPE-V→Spanish assumption is also ASR-confirmed on the audio.

PR #327, opened 2026-08-21
Guard resample overshoot before 16-bit save (stop silent clipping)

wilke0818:
## Summary

The audio-copy sanitize path (`downmix → 16 kHz resample → 16-bit save`) let torchaudio **silently hard-clamp** any sample the resampler pushed past full-scale. That's destructive (flattens peaks) and raises no error. On a source that wasn't already clipped it manufactures clipping — e.g. `diadochokinesis-(v2)-tuh`: source peak 0.998 → resample 1.046 → saved 1.000 with clipping introduced.

Measured on the adult 07_01 build: resample overshoot fired on **~18% of recordings** (all previously silently clamped); **14** of those had clean (un-clipped) sources, so those clamps are unambiguously wrong.

## Fix

`_guard_resample_overshoot(resampled_audio, in_peak)` in `dataset.py`: when the resampled peak exceeds 1.0, rescale the waveform down to the **input peak** so no sample exceeds [-1, 1] — preserving gain and waveform shape, introducing no clipping — and log a warning. No-op when the resampled peak is already in range, so non-overshooting audio is byte-identical to before. Wired into `_copy_audio_files_parallel` right before `save_to_file`.

Rescale-to-input-peak (rather than clamp, or rescale-to-1.0) keeps every file at its own source level whether or not it overshot; overshoot only occurs on near-full-scale material, so the rescale factor is always small.

## Verification

- New `tests/test_resample_clip_guard.py` (2 tests): overshoot → rescaled to input peak with a real 16-bit round-trip showing 0% clipping; in-range → no-op.
- Real case confirmed end-to-end: `diadochokinesis-(v2)-tuh` old-path saved peak 1.000 with clipping, new-path saves 0.998 with 0% clipping.

Independent of the language-aware work (#326); touches only the audio-copy path.

### DIFF
--- a/src/b2aiprep/prepare/dataset.py
+++ b/src/b2aiprep/prepare/dataset.py
@@ -74,11 +74,17 @@ def _guard_resample_overshoot(resampled_audio, in_peak: float):
     resampled peak is already in range. Returns (audio, scale) with scale=None when
     unchanged."""
     wf = resampled_audio.waveform
+    if wf.numel() == 0:
+        return resampled_audio, None
     out_peak = float(wf.abs().max())
     if out_peak <= 1.0:
         return resampled_audio, None
     scale = min(1.0, float(in_peak)) / out_peak
-    return Audio(waveform=wf * scale, sampling_rate=resampled_audio.sampling_rate), scale
+    return (
+        Audio(waveform=wf * scale, sampling_rate=resampled_audio.sampling_rate,
+              metadata=resampled_audio.metadata),
+        scale,
+    )
 
 # Sensitive audio feature content that must not be present for sensitive tasks.
 #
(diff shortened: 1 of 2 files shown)


wilke0818:
**Addressed the code review** (pushed in `3e2ca89`):

- **Carry `Audio.metadata` through the rescale** — it was dropped only on the overshoot branch (asymmetric with the no-op branch).
- **No-op on zero-sample waveforms** — don't call `.max()` on an empty tensor.

**Kept by design:**
- **Global peak-renormalize** (rescale the whole file to the input peak) rather than a local limiter/clamp — clamping is the destructive behavior this PR removes; renormalize preserves waveform shape. Only fires on overshoot (~18% of recordings), and overshoot only occurs on near-full-scale material so the gain change is small.
- The **≤1-ULP floating-point residue** after rescaling an already-full-scale source is below the 16-bit quantization step, so it saves as the intended full-scale sample — no audible/analytic clamp.

Verified against the adult 07_01 build: overshoot fired on ~18% of recordings (all previously silently clamped); 14 had clean sources (the unambiguously-wrong clamps).

PR #328, opened 2026-08-21
Make acoustic_task_id a deterministic UUID keyed by session and task

Evan8456:
Previously acoustic_task_id was just the session folder name, so every acoustic task within a session shared the same id. Derive it instead via uuid5(session, acoustic_task) so each task gets its own stable id, reprocessing the same data never mints a new id, and the Recording -> Acoustic Task join key stays correct.


wilke0818 (approved):
looks correct; LGTM

Issue #329, opened 2026-08-24
Declared dtype schema for packaged dataset artifacts (TSV + parquet)


wilke0818:
## Problem

Every packaged/tabular artifact the dataset emits gets its column **types inferred independently by each writer**, with different engines — pandas for the TSVs, Arrow/HF `Dataset.from_generator` for the parquets. There is no declared schema.

Artifacts affected (at least): `metadata/metadata.tsv`, `features/static_features.tsv`, the feature `*.parquet` (pitch, mfcc, mel_spectrogram, ema, …), `phenotype/*.tsv`, per-subject `sessions.tsv`, `audio_quality_metrics.tsv`.

Consequences of inference-only typing:

- **Lossy/unstable dtypes.** An integer column with any missing value promotes to float (e.g. `stimulus_asset_n_images` → `10.0`); an all-null column becomes object/null; booleans become object.
- **Redcap string-numerics stay strings.** `recording_duration`, `recording_size`, `audio_sample_rate`, `recording_input_gain` ship as text, so every consumer must re-cast.
- **Same column, different dtype across files.** `participant_id` / `session_id` / `task_name` can be typed differently in `metadata.tsv` vs `static_features.tsv` vs a feature parquet → cross-file joins mismatch.
- **No governed missing-value contract.** The "absent JSON key == null cell" semantics were just documented for one field (`stimulus_asset_n_images`) but nothing enforces it globally.
- **The `.json` companions are human description dictionaries, not machine dtype schemas** — pandas/Arrow cannot read them.

## Proposed approach

1. **One field→type registry.** Extend the existing data-dictionary `.json`s with a per-field `dtype` + null policy, or adopt a standard (Frictionless Table Schema / an explicit `pyarrow.Schema`). Shared columns (`participant_id`, `session_id`, `task_name`, ids) get one canonical type used everywhere.
2. **A small coercion utility every writer calls.** Cast to nullable pandas dtypes (`Int64`/`Float64`/`boolean`/`string`) before `to_csv`; build an **explicit** `pyarrow.Schema` for `to_parquet` instead of relying on inference.
3. **Ship the schema beside each artifact** so loaders can apply dtypes on read (TSV carries no embedded types; parquet then carries real ones).
4. **Add a dtype-conformance check to `validate_bundled_dataset`** (this piece also satisfies a checkbox under the "Bundle" section of #319).

## Open design decision

[...]

PR #330, opened 2026-08-24
Capture printed-on-card and multi-image stimulus metadata (Identifying Pictures, Sounds, Story Recall v2)

wilke0818:
## What

Capture stimulus targets that live **on the shown image** for several tasks — previously the sidecars carried only the image URL (or nothing), dropping the actual target.

| task | change |
|---|---|
| **Identifying Pictures** (peds, 37) | printed WORD per card → `speech_type: read`, `stimulus_text` = the word (WER-scorable; several are sight words like "MY"/"GO"/"ONE" that can't be named from the picture), image asset kept |
| **Noisy / Silly / Long Sounds** (peds) | printed sound token per card (`"aa aa aa"`, `"puh puh puh....."`, `"Ahhhhhh... (take a breath)..."`) → stays `non-lexical`, verbatim token as `stimulus_text`, image asset kept |
| **Story Recall v2** (adult) | wordless 10-panel picture story → one **scalar template URL** `StoryRecall_{n}.jpg` + companion **`stimulus_asset_n_images`** count |

## Design points

- **No lists.** The multi-image sequence is a single scalar template URL plus an integer count, so the sidecar stays flat / parquet-friendly (no nested column).
- **`stimulus_asset_n_images`** is present only for sequences; its **absence** is the documented signal that `stimulus_asset` is a single, directly-resolvable image.
- **Format contract documented in `metadata.json`:** an inapplicable optional field is an *absent key* in a JSON sidecar and a *null/None* value in a parquet/TSV column — the two are equivalent.
- **Wordless ⇒ language-independent:** every language uses the same Story Recall panels; only `stimulus_text` differs (Spanish narrative already resolves from the es-419 static bank).
- The bundler casts the count to nullable **`Int64`** so it renders as `10` (not `10.0`) with missing rows as empty cells.

## How the words/tokens were obtained

Transcribed directly from the pinned redcap images (no OCR pipeline) and hand-verified.

## Tests

`tests/test_task_prompts.py` — identifying-pictures read targets, sound-card tokens, Story Recall sequence + `n_images`, single-image tasks omit the count, and a **bundling TSV round-trip** proving absent-key == null-cell and the integer render. **39/39 pass.**

## Follow-up

The `Int64` cast is a stopgap; a general declared-dtype schema for all packaged artifacts (TSV + parquet) is tracked in #329.


wilke0818:
@Evan8456 Mostly standard, does a weird thing with story_recall where we use *_n for the image of the story so we don't have to use lists of URLs since I removed the list of prompts from before though we could return to that if you think this solution is a bit wack

Evan8456:
seems reasonable. I think urls are more troublesome long term, might be the less of 2 evils.

Evan8456 (approved):

PR #333, opened 2026-09-09
fix(redcap2bids): normalize the BIDS task entity at ingest

wilke0818:
## What and why

`redcap2bids` named audio files and sidecars from the raw RedCap recording name with only `.replace(" ", "-").replace("_", "-")`, so case and parentheses reached file names — `task-Audio-Check-(v2)-1`, `task-Diadochokinesis-(v2)-puhtuhkuh`. `deidentify` then rewrote the same entity through `normalize_task_label`, so **the internal and published trees disagreed on every versioned task**, and parenthesised paths are hostile to globs and shell scripts.

Scale in the v4.0.0 exports: 35 of 832 distinct adult recording names carry parentheses (16,725 recordings); peds 11 of 117.

This derives the entity **once**, at ingest, via a new `canonical_task_entity()` = `normalize_task_label()` + a curated alias table. `sanitize_task_entity_in_bids_stem` becomes a verified no-op on a freshly built tree.

## Changes

| Area | Change |
|---|---|
| `utils.py` | `canonical_task_entity()` / `load_recording_name_aliases()`; alias table at `resources/task_registry/recording_name_aliases.json`, curated by hand and kept apart from the generated `registry.json`. A missing resource now raises rather than returning `{}` (which would silently name files differently than a correct install). |
| `dataset.py` | `_write_pydantic_model_to_bids_file` takes the computed entity via `task_entity=`, so the wav and its sidecar cannot be named from two derivations. A recording with no name is skipped with a warning instead of producing `task-nan.wav`. The exclusion list normalizes the task entity on both sides; entries matching nothing are reported. Acoustic-task sidecars gain the collision warning recordings already had. Deidentify's per-file rename warning is aggregated. |
| `fhir_utils.py` | Task-registry resolution goes through the alias table. |
| `prepare.py` | `is_speech_task` normalizes both sides. |

The alias table seeds two families: the six CAPE-V v2 word-order variants (`Cape V sentences-1 (v2)` alongside `Cape V sentences (v2)-1`; the form 3.x published wins), and the bare **`High to Low`** — the Glides pair's second half. All 43 of its sessions also carry `Glides-Low to High` and none carries `Glides-High to Low`.

## Two bugs fixed along the way

- **`is_speech_task` was False for every speech task** under the new naming (`prepare.py:169` compared Title-Case `SPEECH_TASKS` to the file name with a raw substring test) — and was already wrong for `Free speech` under the old naming. Measured 0/7 → 7/7 on sampled tasks; the field is written into every `.pt`.
- **`High to Low` resolved to no registry task at all**, so its 43 recordings shipped without instructions, prompt, or speech_type.

## Behaviour changes to be aware of

- **Every internal-tree file name changes** (lowercasing). Published names are unchanged except the six CAPE-V variants, which now match the dominant form.
- **380 acoustic-task sidecar collisions are now reported.** Two tasks in one session whose names differ only in case (`Free speech` / `Free Speech`) share one sidecar name. Their recordings keep distinct names and both rows remain in `phenotype/task/acoustic_task.tsv` — only the redundant sidecar is lost, and no longer silently. This disappears with the BEP 047 move to one metadata file per recording.

## Verification

Built the full v4.0.0 adult tree (2,005 subjects, 70,520 recordings, 1h39m on 64 cores) plus a peds sidecar-only build:

- **0** task entities contain parentheses or uppercase, in either cohort.
- Counts reconcile: 71,013 export rows − 67 pre-existing audio-check collisions = 70,946 sidecars − 426 failed decodes = 70,520 wavs.
- **The deidentify inclusion list selects the same task set as before** — identical 78 excluded families; no entity kept under v3.1 naming is dropped under the new one.
- **No new within-session recording collisions** in either cohort (the 67 logged are pre-existing audio-check retakes, identical under both namings).
- audio-check exclusion still fires 9/9 adult, 5/5 peds.
- Peds: 117 distinct entities, exactly the set predicted from the export.
- **221 tests pass** (7 new). `tests/test_prepare.py::test_extract_features_timing` errors identically on untouched `origin/main`; `test_cli.py` and `test_generate_audio_features.py` need an installed `b2aiprep-cli`, and `test_update_phenotype_jsons.py` needs `responses` — all pre-existing.

## Follow-up (separate PR)

`audio_filestems_to_remove.json` should key on **recording IDs**, not file stems. 5,363 of its 5,364 entries use reproschema-era `sub-001js` IDs that match nothing in either the v3.1 or v4 peds tree, while the already-present-but-unused `audio_filestems_to_remove_using_recording_ids.json` matches **4,990** recordings in the v4 peds export. That exposure predates this PR and is independent of naming; this PR only ensures no existing exclusion is *weakened*. Worth checking separately whether the shipped v3.1 peds release actually had those removals applied.

### DIFF
--- a/src/b2aiprep/prepare/dataset.py
+++ b/src/b2aiprep/prepare/dataset.py
@@ -99,6 +99,30 @@ _SENSITIVE_FEATURES_REMOVED_FROM_BUNDLE: t.Mapping[str, t.FrozenSet[str]] = {
 }
 
 
+def _note_entity_collision(
+    seen: t.Dict[str, t.Any], entity: str, identifier: t.Any
+) -> t.Tuple[bool, t.Any]:
+    """Record `identifier` under `entity`; report whether a different record already claimed it.
+
+    Returns (collided, prior_identifier). `collided` is True when `entity` has been
+    seen before for a record that is not provably the same one.
+
+    Membership is tested with `in`, never with a truthiness or `is not None` check on
+    the stored value: an id that is missing (None/NaN/"") must still mark the entity as
+    seen. Testing the value instead would read "seen, with a missing id" as "not seen
+    yet" and silently skip the warning for a real collision -- the exact failure these
+    guards exist to catch. Two ids are treated as the same record only when both are
+    present and equal, so NaN != NaN cannot manufacture a false collision either.
+    """
+    if entity not in seen:
+        seen[entity] = identifier
+        return False, None
+    prior = seen[entity]
+    if _is_present(prior) and _is_present(identifier) and prior == identifier:
+        return False, prior
+    return True, prior
+
+
 def _remove_sensitive_features_from_feature_payload(
     features: t.MutableMapping[str, t.Any],
     spec: t.Mapping[str, t.AbstractSet[str]] = _SENSITIVE_FEATURES_REMOVED_FROM_BUNDLE,
@@ -1240,8 +1264,10 @@ class BIDSDataset:
                 acoustic_task_name = acoustic_task_name.replace(" ", "-").replace("_", "-")
                 _task_entity = canonical_task_entity(acoustic_task_name)
                 _task_id = task.get("acoustic_task_id")
-                _prior_task = seen_task_entities.get(_task_entity)
-                if _prior_task is not None and _prior_task != _task_id:
+                _task_collided, _prior_task = _note_entity_collision(
+                    seen_task_entities, _task_entity, _task_id
+                )
+                if _task_collided:
                     _LOGGER.warning(
                         "acoustic_task_name collision: %r maps to the same BIDS task entity "
                         "for participant %s session %s (acoustic_task_id %s and %s); the "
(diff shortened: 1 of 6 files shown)


Evan8456:
## PR #333 Review: `fix/normalize-task-labels-at-ingest` to `release/v4.0.0`

Reviewed against `release/v4.0.0` (confirmed identical to `main`). Overall the normalization approach is solid and well-tested. A couple of things worth addressing before merge.

### Issues

1. **Collision checks can't distinguish "unseen" from "seen with id `None`"**
   `dict.get()` is used in the id-collision checks in `dataset.py`:
   - ~line 1243: acoustic-task collision check
   - ~line 1296: recording-entity collision check

   If a dict ever stores `None` for a missing id, a real collision won't be warned about and a sidecar/audio file will be silently overwritten/dropped. This PR already introduces `fhir_utils._is_present()` to handle the same `None`/`NaN`/empty ambiguity elsewhere, so it's worth reusing it here too so both checks are robust regardless of how a missing id is represented.

2. **Duplicated collision-check logic**
   The two blocks (`~1241-1253` and `~1294-1306`) are near-identical copy-pasted logic. Extracting a shared helper would remove the duplication and make the `_is_present()` fix above a one-place change instead of two.

3. **Minor: redundant computation**
   `_canonical_recording_stem` gets computed twice per path in `_apply_exclusion_list_to_filepaths` (`~lines 2039` and `2054`), which is redundant work that's more noticeable on large datasets.

### What looks good

- Alias routing in `fhir_utils.py`
- `SPEECH_TASKS` normalization fix in `prepare.py`
- `canonical_task_entity` / alias-table loading
- Exclusion-list normalization fix
- Recording-without-a-name guard
- Nice test coverage in `test_task_label_normalization.py`, especially the alias well-formedness and idempotency checks

Evan8456 (approved):
See above, but overall fixes seem minor.

PR #334, opened 2026-09-18
feat(redcap2bids): field map v4, audio-check removal, participant filtering

wilke0818:
## Summary

Make redcap2bids output match what is actually distributed: map every exported column, drop audio checks at ingest, exclude participants without audio, and filter phenotype tables so every row has a corresponding file on disk.

Six commits, each independently testable:

### 1. `c12c582` — ReproSchema fallback for non-dictionary columns

RedCap emits columns that nobody authored as a field (`<form>_complete`, `<form>_timestamp`, structural columns). ReproSchema has no element for them, so mapping one at `delete=NO` crashed the build with `KeyError: 'acoustic_task_complete'`. `_synthetic_data_element()` builds the data dictionary entry from the field map instead. It carries only `description` and `valueType` — deliberately no `termURL`, since no authored definition exists to cite.

Publishing those columns then inflated every table to the full cohort (ALS went 6 → 2005 rows) because RedCap fills `<form>_complete` for every record. `_drop_rows_without_substantive_data()` excludes CSV-described columns from the emptiness test.

### 2. `ddb4368` — Field map: every column mapped, `version_added`, 45 race codes

- **3374 → 3456 rows.** 82 new rows covering the 127 previously-unmapped export columns (56 `_complete`, 22 `_timestamp`, 3 admin, 1 new checkbox option). `only_in_df` drops from 122 → 1 (adult) and 19 → 1 (peds).
- **New `version_added` column.** 872 `v4` / 2584 `pre-v4`, derived from the trusted v3.1 exports. Provides a review filter for newly-arrived fields.
- **45 race option codes corrected.** `asian_race___20297` → `asian_race___2029_7` etc. — `reproschema-py`'s `int("2029_7") == 20297` mangled the underscore. Bijective across 5 parent fields, no other field affected.
- **`ef_researcher_email` → `delete=YES`** — was publishing 277 live email addresses.

### 3. `b6ac9f4` — Audio check dropped at ingest

`is_audio_check()` in utils is the single definition. `_drop_audio_check_rows()` filters the dataframe once before anything reads it, so one filter covers recording.tsv, acoustic_task.tsv, sidecars, and audio copies. The three existing deidentify copies now share the same predicate.

Measured: adult drops 10,062 rows (8,203 Recording + 1,859 Acoustic Task = 11% of recordings), peds 1,668. Registry "no task match" warnings go from 1,668 to 0 in peds — audio check was 100% of them.

### 4. `7922cea` — Exclude participants without audio

Participants whose recordings have no locatable source file get no `sub-*/` directory, no sidecars, and no rows in phenotype. A pre-scan checks source existence and file size (rejects the 259 truncated 4096-byte sources). `recording.tsv` is filtered to only recordings that produced a sidecar.

Measured: adult 2005 → 1527 participants (478 had no audio: 452 enrollment-only + 12 pre-existing empty + 14 audio-check-only). Peds 1079 → 1076.

QA log names every excluded participant and every source-missing recording, with full IDs.

### 5. `a246adc` — `is_calculated` column

Marks the 9 REDCap calculated fields (`@CALCTEXT` / `Field Type = calc`). Enables completeness-based pruning: when a form is Incomplete and only calc fields are populated, the row carries no real data.

### 6. `5f041b7` — Timestamps to `delete=YES`, rename `diagnosis_c_ac`

- **47 timestamp fields → `delete=YES`.** Consumed from the raw dataframe at ingest for derived fields (local hour, session index), never written. The `delete` flag controls phenotype output, not dataframe availability.
- **54 `_complete` fields stay at `delete=NO`.** Needed through to deidentify for completeness-based pruning.
- **`diagnosis_c_ac` → `clinical_global_impression_voice_affecting_condition`** — the REDCap name is opaque; CGI is a recognised clinical assessment framework.

### Invariants now holding

| | adult | peds |
|---|---|---|
| subs == participant.tsv | 1,527 ✓ | 1,076 ✓ |
| recording.tsv == wavs | 62,550 ✓ | 98,341 ✓ |
| audio-check files | 0 | 0 |
| empty sub-* dirs | 0 | 0 |

### Also fixes (pre-existing)

- `data_element.get("question", "").get("en", "")` → `AttributeError` when `question` absent (changed `""` default to `{}`)
- Copy-failure log was truncated to 5 examples; now logs all failed files

### Tests

- `test_csv_only_data_elements.py` — 6 tests for the synthetic data element and row-emptiness rule
- `test_audio_check_filter.py` — 23 tests for `is_audio_check`, the ingest filter, participant exclusion, and a guard that the registry warning is never special-cased
- `test_bids_field_organization.py` — narrowed invariant (unresolved allowed only for RedCap-generated columns) + description check


Evan8456:
## PR #334 Review: `feat/field-organization-v4-columns` → `release/v4.0.0`

Reviewed `6406b28…ea4dba8`: `src/b2aiprep/prepare/dataset.py`, `utils.py`, `bids_field_organization.csv`, and 3 test files. Line numbers refer to `src/b2aiprep/prepare/dataset.py` at PR head unless noted.

The field-organization CSV changes and the row/participant filtering are a clear improvement, and the new tests are good. There are a few issues in the filtering logic that can silently produce a wrong release, which I think should be fixed before merge.

### Blocking

**1. A wrong or empty `--audiodir` now silently produces an empty dataset (`:1513`)**

`audio_files_by_recording` is built as `{}` at `:336` and is never `None`, so every `audio_files_by_recording is not None` guard (`:1513`, `:1634`, `:368`) is unconditionally true.

If `--audiodir` points at a path that doesn't exist, hasn't finished syncing, or whose filenames don't carry the UUID the regex at `:333` expects, then `audio_files_by_recording == {}` → `recordings_with_source` is empty for every participant → `_output_participant_data_to_metadata_file` returns `(False, set())` for all → `participants_with_audio == set()` → `_filter_phenotype_to_participants(phenotype_dir, set())` at `:412` strips every row from every phenotype TSV. The command exits 0 having written no `sub-*/` dirs and header-only phenotype tables.

Before this PR the same mistake just meant "no audio copied". Suggest making the guards truthiness checks, and failing loudly when zero recordings matched rather than emitting an empty release.

**2. The TSV round-trip blanks cells whose literal text is `None`/`NA`/`NULL`/`N/A`/`NaN` (`:927`, `:374`)**

`pd.read_csv(fp, sep="\t", dtype=str)` still applies pandas' default NA tokens — `dtype` only affects parsing of non-NA values. Any cell containing exactly `None`, `NA`, `N/A`, `NULL`, `NaN` or `<NA>` is read as NaN and written back empty by `to_csv`.

Several vendored reproschema items have a choice labelled exactly `None` (e.g. `b2ai-redcap2rs/activities/q_generic_confounders/items/current_neuro_dx`, `ph_standing`, `smoked_substance_in_past_three_months`), and free-text columns can hold the same strings. Since every TSV that loses at least one row gets rewritten, this silently mutates surviving rows. Fix: `keep_default_na=False` (or `na_filter=False`) on both reads.

**3. `record_id` type mismatch can empty every phenotype table (`:389`)**

`participants_with_audio` holds `participant["record_id"]` straight from `pd.read_csv(file_path, low_memory=False, na_values="")` (`redcap.py:726`, no `dtype`), so a numeric-record_id export yields `numpy.int64`. `_filter_phenotype_to_participants` re-reads the TSVs with `dtype=str`, so `df["participant_id"].isin({123})` matches nothing and every table is reduced to its header. It only triggers when at least one participant lacks audio, so it fails intermittently. Normalize both sides to `str`.

**4. `recording.tsv` keeps rows for recordings whose sidecar was never written (`:368`)**

`all_recording_ids_with_sidecar` is set to `recordings_with_source`, which is computed entirely in the pre-scan (`:1484-1510`) and never checks `recording_name`. A recording that has a source file but a missing/NaN `recording_name` is skipped in the main loop at `:1618-1626` (`continue` before the sidecar write) yet stays in `recordings_with_source`, so its `recording.tsv` row survives the filter — exactly the "row referencing a file that was never written" case the filter exists to prevent. The set should be built from recordings that actually reached `_write_pydantic_model_to_bids_file`.

### Non-blocking

**5. The empty-session-dir cleanup can't fire for the case it documents (`:1681`)**

The comment says "This happens when every recording in a session had no source file", but the acoustic-task sidecar is written to `audio_output_path` (~`:1600`) unconditionally, before the recording loop. So when every recording in a session lacks source audio, `ses-*/audio/` still contains the acoustic-task JSONs, `any(session_audio.iterdir())` is truthy, and nothing is removed. Released trees will contain `ses-*/audio/` dirs holding only acoustic-task sidecars and zero audio files. The "empty `sub-*` dirs = 0" invariant currently holds by accident.

**6. Recordings with an empty `recording_id` are dropped with no warning (`:1492`)**

The pre-scan does `if not rec_id: continue`, so an empty-string id lands in neither `recordings_with_source` nor `recordings_without_source`. The main-loop guard at `:1634` then evaluates `"" not in recordings_with_source` → `continue`, so the recording gets no sidecar and no audio copy, and the per-participant QA log at `:1519` never mentions it. Before this PR (and after #333, which deliberately hardened the id-less path) such a recording was still written. At minimum this should be logged.

**7. `sessions.tsv` still lists sessions whose directory was removed (`:1693`)**

`sessions_rows` is accumulated for every session at `:1533` and written unfiltered. A participant with one session that has audio and one that has none (and no acoustic tasks, so the cleanup at `:1681` does fire) gets the empty `ses-*/` removed but keeps its `sessions.tsv` row — the same table/tree inconsistency the participant-level filtering was added to remove.

**8. The new "log every failed file" warning is unbounded (`:206`)**

`_LOGGER.warning("All %d failed audio files … %s", len(failed_files), failed_files)` formats the whole list into one record. On a stale or partially mounted audio share that's tens of thousands of paths in a single multi-megabyte log line; the previous code capped it at 5. Worth capping, or writing the full list to a QA file.

**9. `_MIN_AUDIO_BYTES = 8192` may discard short-but-valid recordings (`:100`)**

8192 bytes is ~0.25 s at 16 kHz/16-bit mono. A genuinely brief recording, or one at a lower sample rate / 8-bit, is treated as a non-recording: it loses its sidecar and audio copy and has its `recording.tsv` row removed, reported only at INFO. The threshold was calibrated against the specific 4096-byte artefacts, so a tighter bound (`> 4096`) or a WAV-header check would be less likely to drop real data.

**10. `_drop_audio_check_rows` mutates the caller's object and returns a `.loc` slice (`:246`)**

`redcap_dataset.df = BIDSDataset._drop_audio_check_rows(...)` writes back onto the caller's `RedCapDataset`, so a caller that reuses it after `from_redcap` silently gets rows removed. Separately, the returned `df.loc[~drop]` slice is assigned into at `:1070` (`df['schema_name_source'] = …`), which raises pandas' `SettingWithCopyWarning`. Returning `df.loc[~drop].copy()` and assigning to a local fixes both.

### Checked and clear

- The 45 `___NNNNN` → `___NNNN_N` race-code renames would break the checkbox choice filter at `:1196` (`str(choice['value']) == str(column_choice)` — the vendored reproschema still has `20297`, giving an empty `choices` list). All 45 rows are `delete=YES`, so that path is never reached. Worth remembering if any of them is ever reactivated.
- `is_audio_check` handles `None` / `""` / `"   "` / NaN / `"nan"` correctly; the `task_name` swap at ~`:2778` is behaviour-preserving for NaN.
- No schema consists solely of `_complete` / `_timestamp` columns, so the `if not substantive: return df` escape hatch in `_drop_rows_without_substantive_data` never keeps a whole-cohort table.
- CSV integrity: no duplicate `(schema_name, column_name)` among the 2900 active rows, no empty `column_name` / `schema_name`, and the counts match the PR body (54 active `_complete`, 0 active `_timestamp`, 9 `is_redcap_calculation=YES`).
- `_construct_phenotype_from_reproschema` auto-prepends `record_id` → `participant_id`, so `task/recording.tsv` and `task/acoustic_task.tsv` *are* reachable by the participant filter.
- `tests/test_bids_field_organization.py` (28 tests) passes against the new CSV. The three `dataset.py`-importing tests could not be run in my environment (`torch` not installed).

wilke0818:
### Additional fix pushed: `34d8440`

**`audio_files_by_recording` None guard** — when `audiodir` is not provided, `audio_files_by_recording` was initialized as `{}` (not `None`), so all the `is not None` guards added by this PR passed on phenotype-only reruns. This caused the pre-scan to classify every recording as missing source audio, and `_filter_phenotype_to_participants` to empty all phenotype tables.

Fix: `audio_files_by_recording` now stays `None` when `audiodir` is not provided, only initialized as `{}` when `audiodir is not None`. All downstream `is not None` checks correctly skip audio-dependent paths on phenotype-only runs.

280/281 non-CLI tests pass (14 CLI failures are pre-existing — `b2aiprep-cli` not on PATH in this env).

wilke0818 (commented):
seems reasonable

src/b2aiprep/prepare/dataset.py
```diff
@@ -833,6 +910,143 @@ def is_only_id_and_demo(row: pd.Series) -> bool:
...
+        that as data keeps a row for every participant in every table. Measured on the v4 adult
+        export, counting them turned `diagnosis/amyotrophic_lateral_sclerosis.tsv` from 6 rows
+        into 2005.
+
+        When every column is bookkeeping there is nothing to test against, so the rows are kept
+        as-is rather than silently emptying the table.
+        """
+        substantive = [c for c in df.columns if c != id_col and c not in csv_only_columns]
+        if not substantive:
+            return df
```
alistairewj: isn't this backwards? the boilerplate says

>  When every column is bookkeeping there is nothing to test against, so the rows are keptas-is rather than silently emptying the table.

which makes no sense to me. if every row is bookkeeping, we should set the entire dataframe to empty right? `df.iloc[0:0]`

alistairewj (approved):

PR #336, opened 2026-09-20
feat(deidentify): rework to per-participant parallelization with allowlist

wilke0818:
## Summary

Reworks the deidentify stage to use per-participant parallelization and an explicit participant allowlist, replacing the file-type-based processing and exclusion list. Stacks on PR #334 (`feat/field-organization-v4-columns`).

- **Participant allowlist**: `participants_to_include.json` is the primary selection mechanism; falls back to inverting `participants_to_remove.json` for backward compatibility
- **Per-participant worker**: each `sub-*/` directory processed as one independent parallel unit (audio, features, sidecars, sessions metadata) via ThreadPoolExecutor, replacing 3 separate file-type passes
- **Session ordinals**: `ses-01`, `ses-02` derived from `session_index` or alphabetical sort, replacing truncated UUIDs
- **Sessions carry-forward**: per-participant `sub-<id>_sessions.tsv` written to deidentified output with disposition-based column dropping, ID remapping, BIDS-compliant naming
- **Disposition column dropping**: phenotype TSVs stripped of `internal` and `review` disposition columns via field map
- **Post-filter phenotype**: phenotype and quality metrics filtered to participants that actually produced output (not the input allowlist)
- **12 false-positive review→release**: structured Likert/checkbox columns reclassified after v3.1 data verification

### Validated against v3.1

Run against `bids_registered_04_14_26` with `shared_release_config`:
- 833/833 participant dirs, 29,289/29,289 wavs, 32,535/32,535 features — exact match
- All 43 phenotype TSVs: identical row counts and participant IDs
- Audio files byte-identical (650 sampled)
- Session dirs: `ses-54ab2a33` → `ses-01` (intentional)
- 12 newly dropped columns all verified as structured data (Likert/checkbox), not free text

### Review fixes applied

- Empty `audio_tasks_to_include` list publishes nothing (was silently publishing everything)
- Sidecar must exist before wav is copied (was copying orphan wavs)
- All-empty output raises RuntimeError
- Exception during per-participant processing cleans up partial output
- Conditional feature stripping: included tasks get `shutil.copy`, excluded tasks get sensitive fields stripped
- Duplicate session_id rows deduplicated before ordinal assignment
- Field map CSV cached after first load (was re-parsed 833+ times)
- `skip_audio=True` phenotype-only runs use allowlist as output set

## Test plan

- [ ] 69 deidentify-related tests pass (22 new in `test_deidentify_rework.py`)
- [ ] End-to-end test: 3-participant allowlist where 1 is emptied by task filtering → phenotype has 2 rows
- [ ] v3.1 real-data comparison validates exact match on all counts
- [ ] Verify `sensitive_audio_tasks.json` is documented as dead config (never loaded by any code path — pre-existing)


wilke0818:
Added `d7079d8` — ported the sidecar wrapper fix from `fix/deid-sidecar-wrapper` (d374b76).

The 2025-09 refactor flattened `{schema: {description, url, data_elements}}` to bare `{column: element}` dicts, so every deidentified sidecar lost its instrument name, description and URL (the published 3.0.0 tree has wrapped sidecars; v3.1 deid/bundle trees are flat).

- Added `load_phenotype_file()` returning `(df, schema_name, header, elements)` for both wrapped and flat (legacy) sidecars
- `load_phenotype_data()` now delegates to it
- `deidentify()` re-wraps on write: `{schema_name: {**header, "data_elements": ...}}`

wilke0818 (commented):
manually reviewed and logic seems to be as intended, minus the session id mapping fallback to alphanumerics which I am removing and changing back to trimming behavior as the fallback

src/b2aiprep/prepare/dataset.py
```diff
@@ -2233,114 +2397,156 @@ def load_audio_tasks_to_include(deidentify_config_dir: Path) -> t.List[str]:
...
+            for pid in participant_allowlist
+            if (self.data_path / f"sub-{pid}").is_dir()
+        )
+        _LOGGER.info("Deidentifying %d participants (max_workers=%d).", len(participant_dirs), max_workers)
+
+        def _process_one(pdir: Path) -> t.Optional[str]:
+            pid = pdir.name[4:]
+            new_pid = participant_ids_to_remap.get(pid, pid)
+            try:
+                had_output = BIDSDataset._deidentify_one_participant(
```
alistairewj: i think `_deidentify_one_participant` is a worse name than `_deidentify_audio_files`, I suggest hybridizing in some way. it's not clear this only operates over audio files

alistairewj (approved):

src/b2aiprep/prepare/dataset.py
wilke0818: well it operates on audio, features, and other sidecar files all at once, deidentifying a participant directory at a time

PR #337, opened 2026-09-21
feat(deidentify): column value review manifest, phenotype parallelization, sensitive-columns migration

wilke0818:
## Summary

Stacks on PR #336 (`feat/deidentify-rework`). Adds per-value review for `review`-disposition columns, parallelizes the phenotype processing bottleneck, and migrates the hardcoded `_remove_sensitive_columns` into the field map.

- **Column value review manifest**: New optional `column_value_reviews.json` config file with per-(participant_id, column_name) verdicts — `safe` (pass through), `redact` (`[REDACTED]`), `drop` (null). Columns with zero verdicts are dropped entirely (backward compat with Phase 1). Verdicts applied before ID remapping so original participant IDs match.
- **Source column name support**: Manifest entries can use either `column_name` (the output/BIDS name) or `source_column_name` (the original REDCap name). Source names are normalized to output names via the field map at load time, so reviewers can work with whichever column naming they have.
- **Phenotype parallelization**: Phenotype TSVs processed in parallel via ThreadPoolExecutor with pre-warmed field map cache. Was the 38-minute bottleneck on v3.1 (serial loop over 43 TSVs).
- **Sensitive-columns migration**: 18 columns reclassified from `release` to correct dispositions — 2 zipcodes → `drop`, 10 surgery dates → `internal` (available for derived fields like days-since-surgery), 6 free-text → `review`. 5 numeric clinical scores stay `release`. `_remove_sensitive_columns` call removed from `_deidentify_phenotype`; gender_identity transform stays inline; method kept deprecated.

### Review fixes included

- Verdict lookup uses pre-remap participant IDs (applied before `_deidentify_phenotype`)
- Empty phenotype dir guarded (`max(1, ...)` on ThreadPoolExecutor workers)
- Verdict values normalized to lowercase at load time
- Normalized inclusion/exclusion sets pre-computed once, passed to per-participant workers

## Test plan

- [ ] Deidentify-related tests pass (including new source_column_name tests)
- [ ] Verdicts: safe passes through, redact replaces, drop nulls, absent nulls, no-verdict column dropped
- [ ] Partial coverage: column kept when some participants have verdicts
- [ ] No manifest: backward-compatible blanket strip of review columns
- [ ] Duplicate verdicts: last wins with warning
- [ ] Source column names normalized to output names via field map
- [ ] Unknown column names (not in field map) kept as-is


wilke0818 (commented):
logic seems mostly right. I think redacted should just be a replacement of what was there (so removed and tagged a PII element) and should have a separate that marks for delete/remove (the else currently handles this)

src/b2aiprep/prepare/dataset.py
```diff
@@ -1007,24 +1007,30 @@ def _build_session_id_mapping(
 
     @staticmethod
     def _drop_columns_by_disposition(
-        df: pd.DataFrame, field_map_df: t.Optional[pd.DataFrame] = None
+        df: pd.DataFrame,
+        field_map_df: t.Optional[pd.DataFrame] = None,
+        keep_review: bool = False,
```
alistairewj: this should just be an enum matching the values and a hierarchy

default > review > internal

default, drop all
review, keep review logic
internal, keep everything

src/b2aiprep/prepare/dataset.py
```diff
@@ -1007,24 +1007,30 @@ def _build_session_id_mapping(
...
     ) -> t.Tuple[pd.DataFrame, t.List[str]]:
         """Drop columns whose disposition is ``internal`` or ``review``.
 
         If *field_map_df* is not supplied it is loaded from the packaged CSV
         (cached after first load).  Falls back to the ``delete`` column when
```
alistairewj: can we just get rid of this fall back? I don't want to accidentally use it.

src/b2aiprep/prepare/dataset.py
```diff
@@ -2066,7 +2180,7 @@ def _warn_about_empty_columns(df: pd.DataFrame, phenotype: dict) -> None:
 
     @staticmethod
     def _remove_sensitive_columns(df: pd.DataFrame, phenotype: dict) -> t.Tuple[pd.DataFrame, dict]:
```
alistairewj, line 2185: just get rid of this

alistairewj (changes requested):

